PR #38, opened 2020-06-30
Added field for policy name

sreyaguha1:
Added policy name


policykit/policyengine/models.py
```diff
@@ -55,6 +55,7 @@ def save(self, *args, **kwargs):
             p.policy_action_code = "action.execute()"
             p.policy_failure_code = "pass"
             p.explanation = "Starter Policy: all policies pass"
+            p.policy_name = "Starter name"
```
amyxzhang, line 58: This is the actual name of a policy. Maybe let's change it to something like "Starter Policy". And the explanation could just be "All policy changes automatically  pass". (Not too important bc we will need to change this in the future).

Issue #6, opened 2020-03-05
get slack username for users who don't sign in to policykit


amyxzhang:
right now I can do actions in slack like postMessage but not have to go to policykit to log in for the first time for it to work. however, this means that policykit doesn't have a SlackUser object for me. We have the user_id from the event handler, so we can create a mostly empty SlackUser object to store the user_id. What would be good to get however is the actual username or readable name (for instance, when postMessage goes through but has to be posted by the Bot account, we can prepend the message with the actual poster's name)

There should be a way to call slack to get that info and populate it at the point where i create that user: https://api.slack.com/methods/users.profile.get

ekim6767:
addressed in pull request 44

PR #44, opened 2020-07-01
creates SlackUsers for all users with IDs filled in upon integration

ekim6767:
admin user still lacks the readable_name attribute, but does have its ID filled in

### DIFF
--- a/policykit/slackintegration/views.py
+++ b/policykit/slackintegration/views.py
@@ -71,25 +71,7 @@ def oauth(request):
             community = None
             user_group,_ = CommunityRole.objects.get_or_create(name="Slack: " + res['team']['name'] + ": Base User")
             user = SlackUser.objects.filter(username=res['authed_user']['id'])
-            
-            if not user.exists():
-                # Checks that user is admin
-                data = parse.urlencode({
-                                       'token': res['access_token'],
-                                       'user': res['authed_user']['id']
-                                       }).encode()
-                reqInfo = urllib.request.Request('https://slack.com/api/users.info', data=data)
-                respInfo = urllib.request.urlopen(reqInfo)
-                resInfo = json.loads(respInfo.read())
-                
-                if resInfo['user']['is_admin'] == False:
-                    response = redirect('/login?error=user_is_not_an_admin')
-                    return response
-                        
-                _ = SlackUser.objects.create(username=res['authed_user']['id'],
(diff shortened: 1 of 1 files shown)


policykit/slackintegration/views.py
```diff
@@ -79,42 +81,54 @@ def oauth(request):
...
+                #get the list of users, create SlackUser object for each user
+                data2 = parse.urlencode({
+                    'token':community.access_token
+                }).encode()
+                                       
+                req2 = urllib.request.Request('https://slack.com/api/users.list', data=data2)
+                resp2 = urllib.request.urlopen(req2)
+                res2 = json.loads(resp2.read().decode('utf-8'))
+                
+                #https://api.slack.com/methods/users.list
```
amyxzhang, line 99: just checking, @ekim6767 - is users.list an admin-only API request? if so, then we don't even need the below call to users.info that checks if it returns an admin error, can just use the result from this call?

ekim6767: I believe that it works with the user, bot and admin tokens, and i'm using the bot token in the code, but I think there is a way to integrate the user.info call with it

amyxzhang: ah well if there's not a way to do check/determination for adminship using users.list then let's forget about itt.

policykit/slackintegration/views.py
```diff
@@ -137,6 +151,22 @@ def is_policykit_action(integration, test_a, test_b, api_name):
 
     return False
 
+#creates slackUser object along with the readable name and user_id
```
amyxzhang, line 154: you should remove this

policykit/slackintegration/views.py
```diff
@@ -178,19 +210,23 @@ def action(request):
                 new_api_action.text = event['text']
                 new_api_action.channel = event['channel']
                 new_api_action.time_stamp = event['ts']
-                u,_ = SlackUser.objects.get_or_create(username=event['user'],
-                                                                           community=community)
+                
+                #u = create_slack_user(community, event['user'])
```
amyxzhang, line 214: same for the rest of the commented out code below

policykit/slackintegration/views.py
ekim6767: this was the start of my solution to #6 which I don't think I got working yet, but I will send a new request without it

Issue #43, opened 2020-07-01
'None' users created


ekim6767:
upon installation of slackintegration, sometimes 'None' users are created in place of admin user

ekim6767:
addressed in pull request 48

Issue #7, opened 2020-03-05
when importing a slack community for the first time grab the policykit Slack bot's bot id


amyxzhang:
need to get and store bot id (user id for a bot user) for the policykit bot user when installing to a new slack community.

then in our listeners, where we try to create new users when we see a user_id we don't know, make sure NOT to create a user if it's policykit's bot user id.

ekim6767:
addressed in pull request 48

PR #48, opened 2020-07-02
code rework for views.py oauth method

ekim6767:
addresses #43, by making sure that all users, including admin that logged in, have their readable_name attribute filled in (no more duplicate Nones created). also prevents from unnecessary SlackUser/SlackCommunity objects being created when a non-admin tries to integrate.

Edit: this integrates Ricky's code for addressing #45 

Update: also adds the bot_id field, addressing #7


amyxzhang:
@ekim6767 looks like there's a conflict due to another PR I merged. Can you resolve it?

Issue #39, opened 2020-06-30
rename process policy everywhere to constitution policy


amyxzhang:

ekim6767:
addressed by pull request 54

PR #49, opened 2020-07-07
Rename

ekim6767:
renames all 'process' with 'constitution' #39


amyxzhang:
can you remove the .DS_Store files? also there's a conflict

PR #56, opened 2020-07-07
Adjusts code to allow for different servers

rickygv99:


amyxzhang:
Some merge conflicts @rickygv99 
Also - could you create a private_template.py file where private.py would normally go with all the variables private.py should name set to empty string?

Issue #61, opened 2020-07-09
permission execute/view/propose need to be implemented


ekim6767:
Right now, these permissions only affect the admin site. Need to actually affect the actions of the users in the site

amyxzhang:
Execute is done. View/propose still needs to be implemented (on admin site and on policyengine).

ekim6767:
implemented in #71

PR #71, opened 2020-07-14
Permissions update

ekim6767:
addresses #61 for slackcommunity actions and constitution actions.


amyxzhang:
I'm going to merge this now but what about the case when a user does *not* have propose permissions on an action? It seems like right now they are still able to propose? I'll make a separate issue for that.

Issue #76, opened 2020-07-15
front end interface for users to select a starter kit


amyxzhang:

ekim6767:
created in PR #94

PR #94, opened 2020-07-23
frontend + deleting unneeded files

ekim6767:
implements the front end for issue #76 after deleting unneeded files


policykit/redditintegration/tasks.py
```diff
@@ -37,7 +37,7 @@ def reddit_listener_actions():
     for community in RedditCommunity.objects.all():
         actions = []
 
-        res = community.make_call('r/policykit/about/unmoderated')
+        res = community.make_call('r/my_cabbages/about/unmoderated')
```
amyxzhang, line 40: we should probably remove references to specific subreddits

policykit/redditintegration/urls.py
```diff
@@ -6,5 +6,6 @@
 urlpatterns = [
     path('oauth', views.oauth),
     path('action', views.action),
-    path('initCommunity', views.initCommunity)
+    path('initRedditCommunity', views.initRedditCommunity),
```
amyxzhang, line 9: I don't think we should use camelcase for URLs

policykit/redditintegration/views.py
```diff
@@ -87,10 +87,22 @@ def oauth(request):
     return response
 
 @csrf_exempt
-def initCommunity(request):
+def getStarterKit(request):
```
amyxzhang, line 90: also generally in python we don't use camelcase. init_community would be more "pythonic"

policykit/slackintegration/views.py
```diff
@@ -143,6 +102,76 @@ def oauth(request):
     response = redirect('/login?success=true')
     return response
 
+@csrf_exempt
+def initSlackCommunity(request):
```
amyxzhang, line 106: one q I have is how much of all this is slack-specific versus something we can do for all integrations at once.

policykit/templates/policyadmin/init_starterkit_slack.html
```diff
@@ -0,0 +1,41 @@
+
```
amyxzhang, line 1: it'd be preferable to have one template for all integrations. this has a lot of code duplication

policykit/policyengine/models.py
```diff
@@ -48,7 +59,7 @@ def save(self, *args, **kwargs):
         if not self.pk:
             super(Community, self).save(*args, **kwargs)
             
-            starterkit = kwargs.get('starterkit')
+            starterkit = self.starterkit
```
amyxzhang, line 62: don't think we should be storing this

PR #103, opened 2020-07-30
V2 dashboard development URL

rickygv99:
Added a variable called VERSION in private.py which, when set to "v2", causes PolicyKit to use the v2 dashboard url instead of the current dashboard url. Allows for easy toggling back and forth between them.

For now, I just copied the home page template into view for the v2 dashboard url -- we will replace this with the v2 dashboard implementation as we implement it.


amyxzhang:
@rickygv99 sorry there are some merge conflicts now

PR #151, opened 2020-08-05
policybundle fix

sreyaguha1:
(do platformpolicybundles and constitutionpolicybundles need receiver functions too? I was verging on yes but wanted to check)

### DIFF
--- a/policykit/policyengine/admin.py
+++ b/policykit/policyengine/admin.py
@@ -158,7 +158,7 @@ admin_site.register(PlatformPolicyBundle, PlatformPolicyBundleAdmin)
 
 
 class ConstitutionPolicyBundleAdmin(admin.ModelAdmin):
-    fields= ('bundled_policies', 'policy_type')
+    fields= ('bundled_policies',)
     #readonly_fields = ('policy_type',)
 
     def save_model(self, request, obj, form, change):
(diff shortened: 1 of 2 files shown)


policykit/policyengine/models.py
```diff
@@ -760,10 +760,6 @@ def save(self, *args, **kwargs):
 
 class PlatformActionBundle(BaseAction):
 
-    bundled_actions = models.ManyToManyField(PlatformAction)
```
amyxzhang, line 763: wait why are we removing this?

policykit/policyengine/admin.py
```diff
@@ -156,7 +158,8 @@ def save_model(self, request, obj, form, change):
 
 
 class ConstitutionPolicyBundleAdmin(admin.ModelAdmin):
-    fields= ('bundled_policies', 'description')
+    fields= ('bundled_policies',)
+    readonly_fields = ('policy_type',)
```
amyxzhang, line 162: i'm confused - policy_type is not a model field right, so why would it show up in the admin page?

policykit/policyengine/models.py
sreyaguha1: In the original, there were two copies of this line so I removed the first. The same text is present on line 771.

PR #163, opened 2020-08-05
Bundlefix (new PR)

sreyaguha1:


policykit/policyengine/models.py
```diff
@@ -760,10 +760,6 @@ def save(self, *args, **kwargs):
 
 class PlatformActionBundle(BaseAction):
 
-    bundled_actions = models.ManyToManyField(PlatformAction)
```
amyxzhang, line 763: again, why are we deleting this field?

policykit/policyengine/admin.py
```diff
@@ -156,7 +158,8 @@ def save_model(self, request, obj, form, change):
 
 
 class ConstitutionPolicyBundleAdmin(admin.ModelAdmin):
-    fields= ('bundled_policies', 'description')
+    fields= ('bundled_policies',)
+
```
amyxzhang, line 162: why are we removing description? i may want to add a description to my bundled policy (also a name)?

policykit/policyengine/models.py
sreyaguha1: This line already exists in the file in line 801 (so this line is a repeat).

policykit/policyengine/admin.py
sreyaguha1: I removed description because models.py did not specify a description field (only a policy_type which was not a model field) so I had assumed that the field description was maybe an error. Should I add both bundle policy name and description as model fields?

amyxzhang: But it's a subclass of a class that does have a description, right? (BasePolicy)?

policykit/policyengine/models.py
amyxzhang: yikes o_o

Issue #164, opened 2020-08-07
permissions for platform actions of starter kits issue


ekim6767:
platform action permissions
- should more account for individual differences in the available platform actions (e.g. do not always give add permissions to all Base Users for all Slack/Reddit actions. There might be some individual platform actions that non-admin Users are not normally able to access in the original community)
- Currently, the save function of the SlackCommunity and RedditCommunity models gives all users the add permissions for all platform actions
-> should integrate the adding of platform permissions into the starterkit initialization process

PR #169, opened 2020-08-10
platform action permissions  bugfix

ekim6767:
- addresses issue #164 
- adds new fields ('constitution_perms', 'slack_perms', 'reddit_perms') so that platform-specific permissions only are added
- gets rid of now unneeded save functions in RedditCommunity and SlackCommunity (that used to add permissions)


policykit/policyengine/models.py
```diff
@@ -216,8 +216,14 @@ class GenericRole(Group):
     is_base_role = models.BooleanField(default=False)
 
     user_group = models.TextField(blank=True, null=True, default='')
-
-    def make_community_role(self, community, creator_token=None):
+    
+    constitution_perms = models.TextField(blank=True, null=True, default='')
```
amyxzhang, line 220: don't think we should be doing this (storing via json, and having specific fields for different types of perms)

PR #189, opened 2020-08-13
discord starter kit integration

### LINKED ISSUE #187: starterkits - discord integration
allow for starter kits to work with discord as well

ekim6767:
resolves #187

### DIFF
--- a/policykit/policyengine/views.py
+++ b/policykit/policyengine/views.py
@@ -172,6 +172,7 @@ def initialize_starterkit(request):
     from django.contrib.auth.models import Permission
     from redditintegration.models import RedditStarterKit
     from slackintegration.models import SlackStarterKit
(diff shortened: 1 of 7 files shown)


policykit/policyengine/views.py
```diff
@@ -183,6 +184,8 @@ def initialize_starterkit(request):
         starter_kit = SlackStarterKit.objects.get(name=starterkit_name)
     elif platform == "reddit":
         starter_kit = RedditStarterKit.objects.get(name=starterkit_name)
+    elif platform == "discord":
+        starter_kit = DiscordStarterKit.objects.get(name=starterkit_name)
```
amyxzhang, line 189: I missed this the last time around - but can you architect it so that we don't need to be calling integration libraries here?

PR #221, opened 2020-08-24
Newsfeed (updated)

sreyaguha1:
I was wondering if you could take a look at my code for this. I have been using django activity stream to create the newsfeed and following the documentation (https://django-activity-stream.readthedocs.io/en/latest/index.html) but am struggling and not sure if I'm missing anything.

Actions are stored in a single table in the database using Django’s ContentType framework and GenericForeignKeys to create associations with different models in your project.


amyxzhang:
can we do post_save()s after more low level classes like BaseAction or BasePolicy to reduce the amt of repetition?

sreyaguha1:
@amyxzhang should i create a model field for the verb to be passed in (as the verb is different for each model)? a
also, i'm not able to successfully render the stream in views.py -- i'm trying to debug this (no luck yet). does the stream need to be created elsewhere?

sreyaguha1:
I was looking at this (https://libraries.io/github/yesbabyyes/django-activity-stream) and was thinking maybe creating the feed needs to be in the models.py itself

sreyaguha1:
I also tried a target stream but no luck -- to clarify, I have been getting a "'WSGIRequest' object has no attribute ..." and have tried to follow StackOverflow posts and github posts but I am unable to decipher what is happening. Could you recommend how to debug this?

amyxzhang:
@sreyaguha1 also some conflicts here. Is this code otherwise ready to go or no?

sreyaguha1:
I fixed conflicts but had some questions/needed help to get newsfeed to work as listed above.

PR #231, opened 2020-08-25
Get total vote count helper function

sreyaguha1:


policykit/policyengine/models.py
```diff
@@ -57,6 +57,11 @@ class Community(PolymorphicModel):
     def notify_action(self, action, policy, users):
         pass
 
+        #   def get_users(self, users):
```
amyxzhang, line 60: Can you remove this commented out code from the PR?

policykit/policyengine/models.py
```diff
@@ -229,6 +234,22 @@ def get_number_votes(self, value=0, users=None):
         else:
             votes = NumberVote.objects.filter(number_value=value, proposal=self)
         return votes
+    
+    def get_total_vote_count(self, vote_type, vote_number = 1, users = None):
```
amyxzhang, line 238: I don't think this should be on the datastore object. shouldn't it be on the proposal object?

sreyaguha1: oops yes sorry!! this was a mistake

amyxzhang:
Hi @sreyaguha1 can you fix the issues and the conflicts?

PR #235, opened 2020-08-27
Dictionary change

sreyaguha1:
Changes helper functions to return dictionaries (as opposed to query sets). This makes it so the user doesn't have to understand query sets to use the helper functions.


amyxzhang:
I don't think we should do this because querysets are really powerful - users can use them to filter/sort/etc. on the database side instead of returning everything and filtering in memory. Instead we have documentation explaining how they can use querysets in their policy.

PR #277, opened 2021-01-21
Fix Slack action permission check

mashton:
The [has_perm function](https://docs.djangoproject.com/en/3.1/ref/contrib/auth/#django.contrib.auth.models.User.has_perm) accepts a string in the format `"<app label>.<permission codename>"`.

Here we were using prefix `slackintegration` (which comes from the PlatformAction `app_name` property) instead of `slack` which is the app label.

I can't get PolicyKit to work without this because the `slackintegration.*` permissions just don't exist. Could this have something to do with how I set it up, or the starter kits? I followed the README instructions to get going, and using the django shell I could see that my User had permissions `slack.add_slackpinmessage` etc (NOT `slackintegration.add_slackpinmessage`).

### DIFF
--- a/policykit/integrations/slack/views.py
+++ b/policykit/integrations/slack/views.py
@@ -241,10 +241,10 @@ def action(request):
                 new_api_action.channel = event['channel_id']
                 new_api_action.timestamp = event['item']['message']['ts']
 
-        if new_api_action.initiator.has_perm('slackintegration.add_' + new_api_action.action_codename):
+        if new_api_action.initiator.has_perm('slack.add_' + new_api_action.action_codename):
             if new_api_action and not policy_kit_action:
                 #if they have execute permission, skip all policies
-                if new_api_action.initiator.has_perm('slackintegration.can_execute_' + new_api_action.action_codename):
+                if new_api_action.initiator.has_perm('slack.can_execute_' + new_api_action.action_codename):
                     new_api_action.execute()
                 else:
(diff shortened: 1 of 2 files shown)


policykit/integrations/slack/views.py
```diff
@@ -241,10 +241,10 @@ def action(request):
                 new_api_action.channel = event['channel_id']
                 new_api_action.timestamp = event['item']['message']['ts']
 
-        if new_api_action.initiator.has_perm('slackintegration.add_' + new_api_action.action_codename):
+        if new_api_action.initiator.has_perm('slack.add_' + new_api_action.action_codename):
             if new_api_action and not policy_kit_action:
                 #if they have execute permission, skip all policies
```
mashton, line 246: @amyxzhang @rickygv99 this code block seems like it is doing the same thing that happens in PlatformAction `save` here: https://github.com/amyxzhang/policykit/blob/c6bcb88240d1bd79e8b01743222ab2fd96883dac/policykit/policyengine/models.py#L779-L816

Is there a reason this permissions check is duplicated in Slack integrations (but not in the other integrations)?

rickygv99: I can't think of any good reason it might be duplicated in just the Slack integration. Since the Slack integration was built before being encapsulated in the more general integration framework, I'm tempted to say it's old purposeless code.

rickygv99 (approved):

PR #279, opened 2021-01-28
Discourse integration

rickygv99:
Still needs enhancement/more testing but bulk of Discourse integration is completed


policykit/integrations/discourse/models.py
```diff
@@ -0,0 +1,266 @@
...
+            data = urllib.parse.urlencode(values).encode('utf-8')
+
+        call_info = team_id + url
+
+        if method:
+            req = urllib.request.Request(call_info, data, method=method)
+        else:
+            req = urllib.request.Request(call_info, data)
+        req.add_header('Content-Type', 'application/x-www-form-urlencoded')
+        req.add_header('User-Api-Key', api_key)
```
mashton, line 55: Does this work? It isn't working for me when I try to curl using the api key from metagov discourse. The [docs](https://docs.discourse.org/) say to use `Api-Key` and `Api-Username` headers, which work for me locally:
```
curl -X GET "https://discourse.metagov.org/directory_items.json?period=all&order=days_visited" -H "Api-Key: [redacted]" -H "Api-Username: system"
```

rickygv99: Tested it out on our local PolicyKit server and it works for me. The auth setup generates API keys automatically for the user rather than requiring them to make one on the Discourse forum. The API keys generated by the setup auth are slightly different from the manually generated ones; they are tagged with user information, so adding in the username is unnecessary for them.

amyxzhang:
I'm going to go ahead and merge this for now. We can address bugs etc. in future PRs.

PR #282, opened 2021-01-28
Ignore Slack events initiated by PolicyKit bot

mashton:
Ignore Slack events where the `user` is the PolicyKit bot. Needed to skip governing when the bot performs a `message` or `pin_added` event. For some reason the `is_policykit_action` function that uses a 2-second heuristic wasn't catching it. I left that check in addition to the new check for bot id. 

Also refactor the action function a bit for clarity, and get rid of NoneType errors that were happening.

### DIFF
--- a/policykit/integrations/slack/views.py
+++ b/policykit/integrations/slack/views.py
@@ -161,8 +161,7 @@ def is_policykit_bot_action(community, event):
 
 def is_policykit_action(integration, test_a, test_b, api_name):
     current_time_minus = datetime.datetime.now() - datetime.timedelta(seconds=2)
-    logs = LogAPICall.objects.filter(proposal_time__gte=current_time_minus,
-                                            call_type=integration.API + api_name)
+    logs = LogAPICall.objects.filter(proposal_time__gte=current_time_minus, call_type=api_name)
 
     if logs.exists():
         for log in logs:
--- a/policykit/integrations/slack/views.py
+++ b/policykit/integrations/slack/views.py
@@ -156,6 +156,9 @@ def oauth(request):
     response = redirect('/login?success=true')
     return response
 
+def is_policykit_bot_action(community, event):
+    return event.get('user') == community.bot_id
+
 def is_policykit_action(integration, test_a, test_b, api_name):
     current_time_minus = datetime.datetime.now() - datetime.timedelta(seconds=2)
     logs = LogAPICall.objects.filter(proposal_time__gte=current_time_minus,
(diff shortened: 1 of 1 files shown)


amyxzhang:
@rickygv99 can you test this out on our policykit server? weird that the 2 second heuristic thing is no longer working?

rickygv99:
Tested it out and it works for me. Not sure why the 2 second heuristic Log API check wasn't working but the fix does solve the problem.

mashton:
Ah it's because the old function is filtering on `LogAPICall.objects.filter(call_type=integration.API + api_name)` which is not turning up any results (see below). I can fix that issue too.
```
In [9]: LogAPICall.objects.filter(call_type='https://slack.com/api/pins.remove')
Out[9]: <QuerySet []>

In [10]: LogAPICall.objects.filter(call_type='pins.remove')
Out[10]: <QuerySet [<LogAPICall: LogAPICall object (1)>, <LogAPICall: LogAPICall object (2)>, <LogAPICall: LogAPICall object (3)>, <LogAPICall: LogAPICall object (4)>, <LogAPICall: LogAPICall object (5)>, <LogAPICall: LogAPICall object (6)>, <LogAPICall: LogAPICall object (7)>, <LogAPICall: LogAPICall object (8)>, <LogAPICall: LogAPICall object (9)>, <LogAPICall: LogAPICall object (10)>, <LogAPICall: LogAPICall object (11)>, <LogAPICall: LogAPICall object (12)>, <LogAPICall: LogAPICall object (13)>, <LogAPICall: LogAPICall object (14)>, <LogAPICall: LogAPICall object (15)>, <LogAPICall: LogAPICall object (16)>, <LogAPICall: LogAPICall object (17)>, <LogAPICall: LogAPICall object (18)>, <LogAPICall: LogAPICall object (19)>, <LogAPICall: LogAPICall object (20)>, '...(remaining elements truncated)...']>
```

Checking for the bot user is more robust than using the 2-second heuristic, though. Do we need both? Are there cases where we need `is_policykit_action` to return true even if the action was not initiated by the PolicyKit bot?

amyxzhang:
Yes there are cases - for instance if PolicyKit is performing an action on behalf of a user (who has logged in to policykit so we have their oauth) instead of as a bot user. One example of this is a Slack channel that can only be posted to after fulfilling some policy---after reverting, at the end, the policy would post the message. Ideally it would just show as coming from the user like a regular Slack channel. 

There some hairy/hacky bits here because for some endpoints, Slack doesn't allow external apps to do things on behalf of logged in users...but basically when we could, we tried to do actions on behalf of the user originating them.

mashton:
Okay I see! Thanks for explaining @amyxzhang. I fixed the previous bug and tested it, so both checks are working now.

amyxzhang:
Sounds good! Thanks!

PR #285, opened 2021-02-12
Expose metagov client for governance processes and resource retrieval

mashton:
Exposes `metagov` client to all steps in policy authoring tool.

A new model, `ExternalProcess`, is introduced to keep track of the outcome of the governance processes. A "governance process" would be something like a vote happening on Loomio. Each `ExternalProcess` record is tied to a `PlatformPolicy` and a `PlatformAction`. When the external process is kicked off, PolicyKit creates an `ExternalProcess` record and uses the primary key to construct a callback URL (`policykit.metagov.org/outcome/:id`). Metagov POSTs the process outcome to that endpoint when the process completes. When PolicyKit receives that request, it updates the `ExternalProcess.outcome` field which is exposed to the policy author through the `metagov.get_process_outcome()` function.

Below are the 3 functions that get exposed to the policy author. (There are policies that use this at https://policykit.metagov.org/main)

1. `metagov.start_process(name, data)` kicks off a governance process

      Policy example:
      ```py
      result = metagov.start_process("loomio", {"closing_at": "2021-05-11"})
      message = f"Poll created at {result.get('poll_url')}, click to vote!"
      ```

2. `metagov.get_process_outcome()` returns the outcome of the process if it has completed

      Policy example:
      ```py
      result = metagov.get_process_outcome()
      if result is None:
          return # still processing
      if result.errors:
          return FAILED
      if result.outcome:
          return PASSED if result.outcome.get('value') == 27 else FAILED
      return FAILED
      ```

3. `metagov.get_resource(name, data)` gets a resource from metagov

      Policy example:
      ```py
      response = metagov.get_resource('cred', { 'username': 'miriam' })
      if response.get('value') > 0.5:
            return PASSED
      ```

##### Issues/concerns that still need to be addressed:
1) Process errors are not handled!
2) Doesn't cover constitution policies or policy bundles
3) PolicyKit cannot yet do any validation or syntax hints for input and output types, because it doesn't know what they are (And neither does metagov core).
4) Even for policies that rely on `get_process_outcome` to evaluate `check` step, the `check` is still being evaluated on a schedule using celery, rather than only being invoked when we receive the outcome for the particular process. This is just unnecessary computation.


policykit/private_template.py
```diff
@@ -1,4 +1,5 @@
-SERVER_URL = ""
+SERVER_URL = 'http://127.0.0.1:3000'
```
rickygv99, line 1: Let's remove these values from the template file

policykit/policyengine/views.py
```diff
@@ -324,13 +325,14 @@ def initialize_policy(policy, action):
     from policyengine.models import Proposal, CommunityUser, BooleanVote, NumberVote
 
     users = CommunityUser.objects.filter(community=policy.community)
+    metagov = MetagovClient(policy, action)
 
     _locals = locals()
     _globals = globals()
 
-    wrapper_start = "def initialize(policy, action):\r\n"
+    wrapper_start = "def initialize(policy, action, metagov):\r\n"
```
rickygv99, line 333: I noticed you expose metagov to all wrappers except for the filter wrapper -- is that intentional?

mashton: Not really, I'll add it there as well.

rickygv99:
looks great!

amyxzhang:
Thank you for writing a test!!

Question - in the interest of potentially reusing code in the future, would it make sense at all to move the metagov client out of the core policyengine code into a clients folder, much like how we have integrations separated out? Say one day we want to have other clients. Also, is clients the right term here?

mashton:
Good question @amyxzhang! Yes I think "client" is the right term -- `metagov.py` is a client interface for accessing the Metagov server's internal endpoints.

It sounds reasonable to decouple the Metagov Client similar to how integrators are separated out, but I will need to think more about what that would look like. (Is there an abstract `BaseClient` and what does it look like? Should `ExternalProcess` continue to be a model defined in the policyengine, or should it be defined in the metagov client package? Should the metagov `/outcome` route be defined in policykit or in the metagov client package? Are we are passing in all available "clients" in to notify/check/pass/etc evaluators? etc..)

To help me answer those questions, can you say more about what the other clients might be? (It feels like there will be overlap here: if a policy author wants to interface with an external system, they could do it by adding a new Metagov plugin, OR by a new "client" to PolicyKit(?)).

amyxzhang:
I see. I guess `metagov.py` is a client according to MetaGov the server, but according to PolicyKit, perhaps it's more of a plugin or a library? One reason I hesitate to call it a client is that in the future, I can imagine different authoring or monitoring views on top of PK if people didn't want to use our default interface. We could expose an API for others for developing authoring clients for instance. All of this terminology is kind of confusing for me, so let me know how you're defining things.

As for decoupling - within our UIST paper, we describe the core driver process but also describe "extensions" (probably not a good name), which implement specific features on top of the core process. This would include roles and permissions and text documents. Some of these should probably just be part of the "core" PK like we have now. The action and policy bundles I think are hard to decouple for instance. Others might be better served as decoupled widgets/plugins that are optional. They could point to external systems (functionality for text documents could point to etherpad for instance) but some could just be implemented internally in PK. The "implemented internally" part would be one reason to decouple metagov as one of a number of "clients". If they're all external then I suppose they *could* just be implemented in MG. But maybe there are reasons why we would want to have an external system directly interfaced in PK vs MG? For instance if we want to incorporate the external system into the PK interface (not just exposed to the policy authors). Another question for decoupling - should PK be able to work without MG (or any other "clients"? My thoughts are fuzzy but I think it should, same as how Loomio, etc. are also standalone?

mashton:
>  Another question for decoupling - should PK be able to work without MG (or any other "clients")?

Yes, I think it should. It already works without MG, and I'm not planning on removing functionality.



Okay, I guess the question then is how should this PR be changed to achieve maximum flexibility to support any of the above possibilities and extensions? Here are 3 potential options. I'm leaning (3) or (2), what do you think?
I'm using the term "client," but I think "library" would make sense too.

1) Merge this as-is, and move metagov out of the core later on, when PK architecture needs to change to support any of the extensions described

2) Restructure as follows
    ```sh
    /clients
    /clients/metagov
    /clients/metagov/models.py # ExternalDecision is defined here
    /clients/metagov/urls.py # The /outcome route is specified here
    /clients/metagov/views.py # The /outcome handler is specified here
    /clients/metagov/client.py # The "client"
    
    /policyengine/views.py # depends on clients.metagov.client, for now
    ```
    For this "client" to be fully decoupled, we would  need to remove the `clients.metagov` dependency from `policyengine`–– instead maybe there is a way for the metagov package to register itself as a client. I think it's too early to implement that at this point, since the future extension points are still vague.

3) I close this PR and keep working off my own fork. I haven't yet added the ability for PolicyKit to react to "trigger" events from Metagov, or for PolicyKit to use Metagov to execute actions on platforms –– and both those pieces of functionality will change the `metagov` interface & how tightly it is coupled to policykit. So I'll probably start with restructuring as mentioned in 2), and develop those new integration points  while keeping in mind that you want metagov to be decoupled–– or loosely coupled–– with PolicyKit core. I can continue to just work on a fork until I arrive at something more stable.

amyxzhang:
I think 2) is fine for now! Let's use "library" instead of "client" - I think it's clearer from the standpoint of a PK developer. 

I think it's fine not to entirely decouple for now.

policykit/integrations/metagov/models.py
```diff
@@ -0,0 +1,17 @@
...
+    policy = models.ForeignKey(PlatformPolicy, on_delete=models.CASCADE)
+    action = models.ForeignKey(PlatformAction, on_delete=models.CASCADE)
+
+    class Meta:
+        unique_together = ['policy', 'action']
+
+# TODO add generic integrator models, so that Metagov can act as a connector for various external platforms
+# ExternalCommunity
+# ExternalUser
+# ExternalPlatformAction]
```
mashton, line 16: Alright @amyxzhang I updated it to move all the metagov-specific views and models into  `integrations/metagov` –– not `clients/metagov` as I initially suggested. Metagov will also be behaving as an "integration" in the sense that it will act as a connector, creating PlatformActions for external platforms (Open Collective, Discourse, etc..). I think it makes sense to keep it all together under `integrations/metagov` rather than splitting out `integrations/metagov` and `libraries|clients/metagov`. These comments highlighted here show a preview of what I'm working on on another branch. I'll have some questions for you & Ricky regarding these generic "ExternalPlatformAction" and such, but I'll ask those in Slack 🙂 

@rickygv99 @amyxzhang  let me know what you think. This expands the meaning of PolicyKit "integrations" to also include "libraries" within them if desired.

amyxzhang:
This looks good (and thank you for tests again T-T). 
Given that your metagov integration deviates a bit from how the other platform integrations are structured, eventually it would be useful to think about whether to standardize the way integrations are structured (as in the different .py files, etc.) and/or have a better separation of the different things that integrations are for. Since I guess we have a broader scope now for them.
But for now, this looks good!

PR #300, opened 2021-03-05
Encapsulate policy execution in "execute_policy" func, and some bug fixes

mashton:
Screen Shot 2021-03-05 at 16 11 22

The logic in the diagram is repeated in 10+ different places in the codebase, and with slight variations and side effects. Policy authors need to rely on a specific sequence of evaluation for policies, so I think we should make this consistent and define it in one place. It will also make debugging a lot easier.

**This moves policy evaluation to a single function called `execute_policy`**.

#### Testing
- [x] Discourse: test a policy that governs creating a topic on Discourse (Note: Discourse integrator doesn't support reverting+Executing actions https://github.com/amyxzhang/policykit/issues/335, but I tested what functionality does exist, and I don't think I'm introducing any new bugs.)
- [x] ~~Discourse: test a policy that votes on a PlatformAction~~ the Discourse integrator doesn't support this yet
- [x] Discord: test a policy that votes on a PlatformAction
- [x] Slack: test a policy that votes on a PlatformAction
- [x] ~~Discourse: test a policy bundle~~ bundles not implemented
- [x] ~~Discord: test a policy bundle~~ bundles not implemented
- [x] ~~Slack: test a policy bundle~~ bundles not implemented

### DIFF
--- a/policykit/integrations/discord/models.py
+++ b/policykit/integrations/discord/models.py
@@ -80,16 +80,17 @@ class DiscordCommunity(Community):
         req.add_header('Content-Type', 'application/x-www-form-urlencoded')
         req.add_header("User-Agent", "Mozilla/5.0") # yes, this is strange. discord requires it when using urllib for some weird reason
 
-        res = None
         try:
             resp = urllib.request.urlopen(req)
-            res = json.loads(resp.read().decode('utf-8'))
         except urllib.error.HTTPError as e:
             logger.info('reached HTTPError')
             logger.info(e.code)
             raise
 
-        return res
+        res = resp.read().decode('utf-8')
+        if res:
+            return json.loads(res)
+        return None
 
     def execute_platform_action(self, action, delete_policykit_post=True):
         from policyengine.models import LogAPICall, CommunityUser
--- a/policykit/integrations/discourse/tasks.py
+++ b/policykit/integrations/discourse/tasks.py
@@ -55,7 +55,7 @@ def discourse_listener_actions():
             username = usernames[0]
             call_type = '/posts.json'
             if not is_policykit_action(community, call_type, topic, username):
-                t = DiscourseCreateTopic.objects.filter(id=topic['id'])
+                t = DiscourseCreateTopic.objects.filter(community=community, topic_id=topic['id'])
                 if not t.exists():
                     logger.info(f"[celery-discourse] creating new DiscourseCreateTopic object for topic {topic['title']}")
 
@@ -72,7 +72,7 @@ def discourse_listener_actions():
                     new_api_action.title = topic['title']
                     new_api_action.category = topic['category_id']
                     new_api_action.raw = raw
-                    new_api_action.id = topic['id']
+                    new_api_action.topic_id = topic['id']
 
                     u,_ = DiscourseUser.objects.get_or_create(
                         username=username,
--- a/policykit/integrations/discourse/tasks.py
+++ b/policykit/integrations/discourse/tasks.py
@@ -4,7 +4,6 @@ from celery import shared_task
 from celery.schedules import crontab
 from policyengine.models import Proposal, LogAPICall, PlatformPolicy, PlatformAction, BooleanVote, NumberVote
 from integrations.discourse.models import DiscourseCommunity, DiscourseUser, DiscourseCreateTopic, DiscourseCreatePost
-from policyengine.views import filter_policy, check_policy, initialize_policy
 from urllib import parse
 import urllib.request
 import urllib.error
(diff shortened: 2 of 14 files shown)


policykit/policyengine/views.py
```diff
@@ -658,3 +659,51 @@ def document_action_remove(request):
     action.save()
 
     return HttpResponse()
+
+def execute_policy(policy, action):
+    """
+    Execute policy for given action. This can be run repeatedly to check proposed actions.
+    """
```
mashton, line 666: The idea is that `filter_policy`/`check_policy`/etc functions are never called outside of this function.

policykit/policyengine/views.py
```diff
@@ -438,6 +438,7 @@ def clean_up_proposals(action, executed):
             p.save()
 
     p = action.proposal
+    # what is going on here
```
mashton, line 441: Can we remove the `clean_up_proposals` function in favor of:
* After running the `pass_policy` block, always mark the action.proposal as `PASSED`. If it is a bundle, also mark the bundle as `PASSED`.
* After running the `fail_policy` block, always mark the action.proposal as `FAILED`. If it is a bundle, also mark the bundle as `FAILED`.
?

amyxzhang:
Overall, I think the refactor to move all policy execution into one function sounds good!

The slight variations are due to where it's being executed, which could just be passed in to the function/set on the action object. Actions can be attempted both on the platform and on the PK website (that's the `community_origin` variable). They can be bundled or not. `community_revert` sets whether the action was reverted previously.

The way bundles work is a little tricky. An action bundle is both a subclass of `BaseAction` (so is an action) and also contains a list of actions. So since we attach proposals to `BaseAction`, both the bundle itself as a whole has a proposal attached to it, as does each action within the bundle. This allows us to model an `ELECTION` like so: the election "passes" as in it completes successful, so the bundle's proposal is set to `PASSED`. Within the specific actions in the bundle though, one of them pass, while the rest fail. (Since it's an election, there's only 1 winner.) This is different from a bundle of type `BUNDLE` (I know, sorry), where passing the bundle means passing every action within the bundle.

policykit/policyengine/models.py
```diff
@@ -836,29 +810,7 @@ def save(self, *args, **kwargs):
...
-
-                                check_result = check_policy(policy, action)
-
-                                logger.info('just checked policy: ' + policy.name)
-                                if check_result == Proposal.PASSED:
-                                    pass_policy(policy, action)
-                                elif check_result == Proposal.FAILED:
-                                    fail_policy(policy, action)
-                                    if self.community_origin:
-                                        self.community_revert = True
```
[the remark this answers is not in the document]
mashton: Issue for this lifting `revert` and `execute` into policy: https://github.com/amyxzhang/policykit/issues/318

policykit/integrations/discourse/tasks.py
```diff
@@ -114,17 +113,3 @@ def discourse_listener_actions():
...
-                logger.info('about to filter policy')
-                if filter_policy(policy, proposed_action):
-                    logger.info('just filtered policy')
-                    cond_result = check_policy(policy, proposed_action)
-                    if cond_result == Proposal.PASSED:
-                        logger.info('passed proposal discourse')
-                        pass_policy(policy, proposed_action)
-                    elif cond_result == Proposal.FAILED:
-                        logger.info('failed proposal discourse')
-                        fail_policy(policy, proposed_action)
```
mashton, line 130: @rickygv99 I think this can be removed entirely because the celery task at `policyengine/tasks.py` already re-evaluates all `proposed` PlatformActions. Am I missing something?

policykit/integrations/discord/models.py
```diff
@@ -80,16 +80,17 @@ def make_call(self, url, values=None, action=None, method=None):
         req.add_header('Content-Type', 'application/x-www-form-urlencoded')
         req.add_header("User-Agent", "Mozilla/5.0") # yes, this is strange. discord requires it when using urllib for some weird reason
 
-        res = None
         try:
             resp = urllib.request.urlopen(req)
-            res = json.loads(resp.read().decode('utf-8'))
```
mashton, line 86: This was throwing for `DELETE` calls because there is no response body

policykit/policyengine/views.py
mashton: leaving as is, created https://github.com/amyxzhang/policykit/issues/342

policykit/integrations/discourse/tasks.py
rickygv99: Yes, I think we should be okay to remove it!

rickygv99 (approved):

Issue #299, opened 2021-03-02
"UNIQUE constraint failed" when the same username is used on two platforms


mashton:
If there is a username that is duplicated across platforms (eg `miriam123` on reddit and `miriam123` on discourse), PolicyKit crashes with `django.db.utils.IntegrityError: UNIQUE constraint failed: auth_user.username` because the `CommunityUser.username` field is supposed to be unique. 

Possible fix:
Update user model to be unique on `username` _and_ `community`. See relevant [django docs on specifying a custom user model](https://docs.djangoproject.com/en/3.1/topics/auth/customizing/#specifying-a-custom-user-model).

@rickygv99 @amyxzhang have been discussing how to support multiple platforms simultaneously, including support for cross-platform policies --  maybe this is part of that larger discussion.

mashton:
@shaunagm hit this recently:

Community A has a `OpencollectiveCommunity` for the "metagov" collective. It is linked to an `OpencollectiveUser` with usernama `shauna`.

Community B has a `OpencollectiveCommunity` that is ALSO linked to the "metagov" collective. When Shauna performs an action on OC, the handler tries to `get_or_create` an OpenCollectiveUser with username "shauna" linked to Community B, which throws an exception because the username `shauna` is not unique. https://github.com/amyxzhang/policykit/blob/623523d76d63c06b6d559ad7b477d80512fbd2e7/policykit/integrations/opencollective/handlers.py#L58-L62

mashton:
Once this is fixed, we should get rid of the hack in the Discord integration where the username gets prefixed by the guild ID

PR #302, opened 2021-03-12
Support policies that are triggered by Metagov events

mashton:
With this change, a Policy author can specify an event defined in a Metagov plugin as a policy trigger. Example `filter` block: 

    ```
    return action.action_codename == 'metagovaction' and action.event_type == 'opencollective_expense_created'
    ```

#### How this works:

> Pre-requirement: there is a SlackCommunity (or whatever Community) that has already been configured to enable some Metagov plugin that sends events.

1. Metagov makes a POST request to the PolicyKit endpoint `/metagov/action` with a payload that looks like:
   ```js
       {
            "community": "slack-test123",   # Unique string identifying the community that this action "belongs" to. It was originally defined by PolicyKit.
            "initiator": {"user_id": "miriam", "provider": "discourse"},
            "event_type": "discourse_post_created",
            "data": {"title": "test", "raw": "post text", "category": 0},
        }
   ```
2. The `integrations.metagov` app in PolicyKit creates a `MetagovPlatformAction` for this event.
    * The `action.initiator` is set to a `MetagovUser` which has blanket access to "propose" any MetagovPlatformAction in this community.
    * The `action.community` is set to the _original community in PolicyKit_ (in this example, the `SlackCommunity`).

    > The `MetagovPlatformAction` does not support `execute` or `revert` functionality. It is simply a trigger event. In order to support execution/reversion of actions, we'd need to create individual subclasses of MetagovPlatformAction for each action (eg `OpenCollectiveConversation`, `DiscoursePost`, etc.)
3. When the newly created `MetagovPlatformAction` event is saved, it triggers a policy evaluation. The [logic is](https://github.com/amyxzhang/policykit/blob/c5569bf20aec47c9f69176a536d0e23879308beb/policykit/policyengine/models.py#L799) to execute all policies where `PlatformPolicy.objects.filter(community=action.community)`. That means any PlatformPolicy that is defined for this SlackCommunity will be tested.
    > That means Slack policies with `filter: return True` will execute for metagov actions too, which is maybe not what we want? I don't know how those blanket policies are really used.

### DIFF
--- a/policykit/integrations/metagov/models.py
+++ b/policykit/integrations/metagov/models.py
@@ -1,5 +1,13 @@
+import json
+import logging
+
+from django.contrib.auth.models import ContentType, Permission, User
 from django.db import models
-from policyengine.models import PlatformAction, PlatformPolicy
+from policyengine.models import (Community, CommunityRole, CommunityUser,
+                                 ConstitutionPolicy, PlatformAction,
+                                 PlatformPolicy, Proposal, StarterKit)
+
+logger = logging.getLogger(__name__)
 
 
 class ExternalProcess(models.Model):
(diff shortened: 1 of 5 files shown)


policykit/integrations/metagov/models.py
```diff
@@ -10,8 +15,57 @@ class ExternalProcess(models.Model):
...
-# TODO add generic integrator models, so that Metagov can act as a connector for various external platforms
-# ExternalCommunity
-# ExternalUser
-# ExternalPlatformAction]
+class MetagovUser(CommunityUser):
+    """
+    Represents a user in the MetagovCommunity, which could be on any platform.
+    """
+    # Hack so it doesn't clash with usernames from other communities (django User requires unique username).
+    # TODO(#299): make the CommunityUser model unique on community+username, not just username.
```
mashton, line 23: https://github.com/amyxzhang/policykit/issues/299

policykit/policyengine/models.py
```diff
@@ -815,6 +815,11 @@ def pass_action(self):
         proposal.save()
         action.send(self, verb='was passed')
 
+    def get_applicable_policies(self):
+        if self.app_name == 'metagov':
+            return PlatformPolicy.objects.all()
+        return PlatformPolicy.objects.filter(community=self.community)
```
mashton, line 821: I need to rethink this actually, because I didn't consider cases where a single PolicyKit instance is serving multiple communities.

amyxzhang:
@mashton my bad for reviewing out of order....could you resolve the merge conflict? Sorry about that!

amyxzhang:
Hi @mashton does this still need to be reviewed as is or should I be waiting for you to push more commits?

mashton:
I need to update it since the most recent merge, I’ll ping you when it’s
ready for review again!

amyxzhang:
This looks good! We'll want to add PolicyKit documentation around these action codenames and event types (once we have a readthedocs page for how to write policies for the different integrations) and eventually surface them in the policy editor too.

PR #304, opened 2021-03-16
Support closing a metagov process from a policy

mashton:
Let policy authors "close" an existing governance process by invoking `DELETE /api/internal/process/<process_name>/<process_id>`.

The plugin author implements "close" if they want to expose a way for Metagov/Driver to close or cancel a governance process manually. In some cases it will not be implemented, because the plugin just notifies the Driver when the process is completed––without giving Metagov a way to cancel it manually (Loomio plugin works like this).

In the `discourse-poll` process I'm using,  the "close" function closes the poll in discourse, locks the post, and returns the vote outcome. Here's an example usage of that process from a Platform Policy:

```py
# NOTIFY
parameters = {"title": "Should this post be pinned?", "options": ["yes", "no"], "category": 8}
result = metagov.start_process("discourse-poll", parameters)
poll_url = result.get('poll_url')
# Elided: notify users of the poll
```
```py
# CHECK
import datetime
if action.proposal.time_elapsed() < datetime.timedelta(minutes=5):
  return None #still processing

result = metagov.close_process()

if result.errors:
  return FAILED
if result.outcome:
  agree_count = result.outcome.get('yes')
  disagree_count = result.outcome.get('no')
  return PASSED if agree_count > disagree_count else FAILED

return FAILED
```

### DIFF
--- a/policykit/integrations/metagov/library.py
+++ b/policykit/integrations/metagov/library.py
@@ -30,18 +30,18 @@ class Metagov:
             policy=self.policy,
             action=self.action
         )
-
         url = f"{settings.METAGOV_URL}/api/internal/process/{process_name}"
         payload['callback_url'] = f"{settings.SERVER_URL}/metagov/outcome/{model.pk}"
         logger.info(f"Making request to start '{process_name}' with payload: {payload}")
         response = requests.post(url, json=payload)
         if not response.ok:
-            logger.error(f"Error starting process: {response.status_code} {response.text}")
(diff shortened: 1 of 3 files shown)


amyxzhang:
Looks good! Not urgent but it would be great to get documentation into our readthedocs on this library for policy authors.

amyxzhang:
@rickygv99 would you be able to migrate our db after merging these changes on our prod server?

PR #324, opened 2021-04-09
Add PolicyKit settings page, include Metagov config in it

mashton:
Add a settings page at https://policykit.metagov.org/main/settings/ and link to it from the header.
The Metagov configuration UI from https://github.com/amyxzhang/policykit/pull/311 is `include`d in the settings template. The settings page can be extended to add more PolicyKit specific settings at any time.
Only community admins are allowed to view the Metagov config. If the user is not an admin they see a message like `"Metagov is enabled for this community, but you do not have access to view the configuration"`.


policykit/policykit/settings.py
```diff
@@ -37,6 +37,7 @@
                  'ec2-54-190-197-117.us-west-2.compute.amazonaws.com',
                  '54.190.197.117']
 
+METAGOV_ENABLED = True if METAGOV_URL else False
```
mashton, line 40: @rickygv99 @amyxzhang also adding this `METAGOV_ENABLED` flag in settings, and making sure we check it anywhere we do Metagov things in policy engine. This ensures that PolicyKit still works without Metagov too. (I had introduced a bug that @rickygv99 caught with his recent PR, thansk! this should make it easier going forward)

PR #333, opened 2021-04-12
Set is_community_admin for Discourse users; use it to set new can_edit_metagov_config perm

mashton:
Fixes some bugs in DiscourseUser creation:
* Set `is_community_admin` if user is admin. This gets set when creating all users on community install, and gets updated when user logs in.
* Set `avatar` to correct link

### DIFF
--- a/policykit/policyengine/views.py
+++ b/policykit/policyengine/views.py
@@ -130,7 +130,7 @@ def settings_page(request):
     user = get_user(request)
     community = user.community
 
-    if METAGOV_ENABLED and user.is_community_admin:
+    if user.has_perm("metagov.can_edit_metagov_config"):
         metagov_config = get_or_create_metagov_community(community)
         if metagov_config:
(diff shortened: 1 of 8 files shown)


policykit/policyengine/views.py
```diff
@@ -130,7 +130,7 @@ def settings_page(request):
     user = get_user(request)
     community = user.community
 
-    if METAGOV_ENABLED and user.is_community_admin:
+    if user.has_perm("metagov.can_edit_metagov_config"):
```
mashton, line 133: @amyxzhang I updated this PR as you suggested to create a custom Role (`"Metagov Admin"`) with a single permission (`"can_edit_metagov_config"`). right now users with `is_community_admin` get automatically assigned to this role. 

Screen Shot 2021-04-12 at 16 02 56

amyxzhang: Looks great!

Issue #332, opened 2021-04-12
Discourse integration uses non-unique topic ID as primary key for PlatformActions


mashton:
This will break the primary key uniqueness constraint when we have more than 1 active discourse community, because topic IDs will be repeated. 
https://github.com/amyxzhang/policykit/blob/a023c7be64e84ad6f95439dba1d7d57c3ef543ef/policykit/integrations/discourse/tasks.py#L66

@amyxzhang @rickygv99

mashton:
We also set `id` like this, I'm not sure how this is supposed to work:

https://github.com/amyxzhang/policykit/blob/a405abfbca550d60e8db1bcee55f995662ce951f/policykit/integrations/discourse/models.py#L128

mashton:
Fixed in https://github.com/amyxzhang/policykit/pull/300/files, the Discord integration has the same problem though. I'll fix on that branch too.

mashton:
fixed in https://github.com/amyxzhang/policykit/pull/300

PR #344, opened 2021-04-16
Post loading

### LINKED ISSUE #319: Discourse integration loads all posts on startup, gets rate limited
When installing PolicyKit to a Discourse server for the first time, it fetches all existing posts individually to "govern" them, so it's hammering the Discouse API on install,which immediately leads to rate limiting.

Suggestion: when PolicyKit is installed to a new platform, I don't think it should try to retroactively "govern" all the actions that have already taken place.

rickygv99:
Fixes #319 

There are two possible ways to fix the rate-limiting problem / Discourse integration loading on startup all previous posts that occurred prior to its integration into the community.
 1. Compare the post creation time to the PK community creation time. Only create an object in the db for the post if post creation time is after PK community creation time.
 2. Compare the post creation time to the current time. Only create an object in the db if the post occurred recently enough to the current time (within the last two minutes).

Ultimately I decided on approach 2 to avoid cases where PK may be added/removed to the community multiple times (we don't want to retroactively govern posts that occurred during periods where PK was removed). Let me know if you would prefer approach 1 though!

### DIFF
--- a/policykit/integrations/discourse/tasks.py
+++ b/policykit/integrations/discourse/tasks.py
@@ -15,21 +15,22 @@ import json
 
 logger = logging.getLogger(__name__)
 
-def is_policykit_action(community, call_type, topic, username):
-    if username == 'PolicyKit': ## TODO: Compare IDs in future, not usernames
-        return True
+def should_create_action(community, call_type, topic, username):
+    # If topic already has an object, don't create a new object for it.
+    if DiscourseCreateTopic.objects.filter(id=topic['id']).exists():
+        return False
 
-    current_time_minus = datetime.datetime.now() - datetime.timedelta(minutes=2)
-    logs = LogAPICall.objects.filter(
-        proposal_time__gte=current_time_minus,
-        call_type=call_type
-    )
-    if logs.exists():
-        for log in logs:
-            j_info = json.loads(log.extra_info)
-            if topic['id'] == j_info['id']:
-                return True
-    return False
+    created_at = topic['created_at']
+    created_at = created_at.replace("Z", "+00:00")
+    created_at = datetime.datetime.fromisoformat(created_at)
+
+    # If topic is more than two minutes old, don't create an object for it.
+    # This way, we only create objects for topics created after PolicyKit
+    # has been installed to the community.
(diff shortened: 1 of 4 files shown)


policykit/integrations/discourse/tasks.py
```diff
@@ -15,21 +15,25 @@
...
-    if logs.exists():
-        for log in logs:
-            j_info = json.loads(log.extra_info)
-            if topic['id'] == j_info['id']:
-                return True
-    return False
+def should_create_action(community, call_type, topic, username):
+    # If topic already has an object, don't create a new object for it.
+    if DiscourseCreateTopic.objects.filter(id=topic['id']).exists():
+        return False
```
mashton, line 21: this will have a merge conflict with https://github.com/amyxzhang/policykit/pull/300 since it creates the field `topic_id` (because https://github.com/amyxzhang/policykit/issues/332). just a heads up whoever merges in second will need to change it

policykit/integrations/discourse/tasks.py
```diff
@@ -15,21 +15,25 @@
...
+    created_at = datetime.datetime.fromisoformat(created_at)
+
+    now = datetime.datetime.now()
+    now = now.replace(tzinfo=datetime.timezone.utc) # Makes the datetime object timezone-aware
+
+    # If topic is more than two minutes old, don't create an object for it.
+    # This way, we only create objects for topics created after PolicyKit
+    # has been installed to the community.
+    if now - created_at > datetime.timedelta(minutes=2):
+        return False
```
mashton, line 34: This seems like a reasonable approach, but it has a dependency on this value, right?
https://github.com/amyxzhang/policykit/blob/ea3ac7e49078daade578726ec348e0f4c35d27f5/policykit/policykit/settings.py#L238-L239

If the beat schedule is >2m, then we'll miss some posts. maybe add a comment next to each value saying that they are dependent

rickygv99: Updated the code -- now the Celery beat schedule time is stored as a constant (CELERY_BEAT_FREQUENCY) in settings.py. discord/tasks.py checks if the topic is more than 2 * CELERY_BEAT_FREQUENCY seconds old and only creates objects for topics that have been created less than that many seconds ago.

PR #348, opened 2021-04-19
Only execute one policy per action, choose the most recently modified policy

mashton:
I'll describe this change through an example. Assume that action X passes the `filter` block for policies A and B, and fails the `filter` block for policy C. Policy A was created before policy B.

**Old behavior:**

When action X occurs, both policies A and B get fully executed (the full sequence of check/notify/pass/fail occurs for both policies). Policy A gets evaluated first, then policy B gets evaluated. Policy A decides that the action failed, it runs the policy's `fail` code and sets the proposal status to `failed`. Then policy B decides that the action passed, and runs policy B's `pass` code and sets the proposal status to `passed. And so on. The interactions between the policies is unclear to the author.

**New behavior:**

When action X occurs, only policy B gets fully executed (the full sequence of check/notify/pass/fail). Policy A is ignored, because it is older.

**Bugs:**

There is still some unpredictable behavior for actions that are in "proposed" state waiting for a vote to occur (or other async process). If policies are modified during the wait-time, the action might suddenly pass/fail due to some _other_ policy (NOT the one that initiated the voting process). This is not a _new_ bug, but it is a bug that will continue to exist. I think the desired behavior is that when an action moves into `proposed` state and starts getting polled every minute, it should always be polled against the same Policy that originally caused it to go into `proposed` state. This requires a link between the `action` and the `policy`, which doesn't currently exist.

### DIFF
--- a/policykit/integrations/discourse/models.py
+++ b/policykit/integrations/discourse/models.py
@@ -52,8 +52,10 @@ class DiscourseCommunity(Community):
         try:
             resp = urllib.request.urlopen(req)
         except urllib.error.HTTPError as e:
-            logger.info('reached HTTPError')
-            logger.info(e.code)
+            logger.error('reached HTTPError')
+            logger.error(e.code)
+            error_message = e.read()
+            logger.error(error_message)
             raise
 
         resp_body = resp.read().decode('utf-8')
--- a/policykit/integrations/discourse/models.py
+++ b/policykit/integrations/discourse/models.py
@@ -182,7 +182,7 @@ class DiscourseCreatePost(PlatformAction):
     def execute(self):
         # only execute the action if it didnt originate in the community, OR if it was previously reverted
         if not self.community_origin or (self.community_origin and self.community_revert):
-            reply = self.community.make_call('/posts.json', {'raw': self.raw})
+            reply = self.community.make_call('/posts.json', {'raw': self.raw}) #FIXME this needs to have topic_id
             self.post_id = reply['id']
             self.save()
         super().pass_action()
--- a/policykit/integrations/discourse/models.py
+++ b/policykit/integrations/discourse/models.py
@@ -150,7 +150,8 @@ class DiscourseCreateTopic(PlatformAction):
         super().revert(values, call, method='DELETE')
 
     def execute(self):
-        if not self.community_revert:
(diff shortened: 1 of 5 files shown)


policykit/integrations/discourse/models.py
```diff
@@ -150,7 +152,8 @@ def revert(self):
         super().revert(values, call, method='DELETE')
 
     def execute(self):
-        if not self.community_revert:
+        # only execute the action if it didnt originate in the community, OR if it was previously reverted
+        if not self.community_origin or (self.community_origin and self.community_revert):
```
mashton, line 156: https://metagov.slack.com/archives/C01FSJD9W3Z/p1618933470276700?thread_ts=1618928075.273600&cid=C01FSJD9W3Z

amyxzhang:
>  I think the desired behavior is that when an action moves into proposed state and starts getting polled every minute, it should always be polled against the same Policy that originally caused it to go into proposed state. This requires a link between the action and the policy, which doesn't currently exist.

Hm I see, so you see it as like the "filter" step is really more of an "assign" step, where a proposed action get assigned to one (or more - I think this is important) policies during its first pass over all the policies, after which it doesn't need to go through them again. I think this way of doing it certainly simplifies some things.

mashton:
Yeah exactly, an action would get "assigned" to one or more policies. Right now we have this implicit link between actions and policies, with the "proposal" object: `action.proposal.status="proposed"` _implies_ a link to some specific polic(ies), but there's no way of knowing which policies. Having an explicit link between actions and policies might also make it easier to build monitoring tools– since you'd be able to tell from the data model exactly  _why_ a particular action is pending or why it passed/failed. Right now we rely on logging to get this information.
Agreed it would simplify things, but it also restricts certain things. I'm sure there are reasons against this approach that I'm not thinking of!

amyxzhang:
The main complication I can think of is what to do when policies change while an action is still in proposed state. Like if that assigned policy were abolished or edited, or another policy is suddenly in scope. Right now, everything is evaluated according to what policies are "live". If we assign policies to an action, in a sense it is capturing a snapshot of a particular state and writing it down. But maybe you actually wouldn't want to have everything evaluated live, since maybe old actions *should* be evaluated according to the policy state when they were proposed. 

I think everything else is actually the same, so long as we allow for multiple policies to be assigned to an action (in the future, not now).

PR #351, opened 2021-04-22
Fix discourse 422 bug

rickygv99:
Fixes the 422 error which Discourse throws when attempting to recover a topic that has already been deleted. Error is thrown because re-making the post rather than recovering it triggers Discourse's "Topic name has already been used" filter.

### DIFF
--- a/policykit/scripts/starterkits.py
+++ b/policykit/scripts/starterkits.py
@@ -8,7 +8,7 @@ from integrations.slack.models import SlackStarterKit
 from policyengine.models import *
 from polymorphic.models import PolymorphicModel
 
-# default starterkit -- all users have ability to view/propose actions + all actions pass automatically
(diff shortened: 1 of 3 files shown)


policykit/scripts/starterkits.py
```diff
@@ -30,9 +30,18 @@
...
+    "notify": "pass",
+    "success": "action.execute()",
+    "fail": "pass",
+}
+
 testing_policy1_slack = GenericPolicy.objects.create(
-    **all_actions_pass,
-    description="Starter Constitution Policy: all constitution actions pass automatically",
+    **all_actions_execute,
+    description="Starter Constitution Policy: all constitution actions execute automatically",
```
mashton, line 44: @rickygv99 can you explain why these starterkit policies need to call `action.execute()`?

PR #352, opened 2021-04-26
Make Discord Integration fully-functional

rickygv99:
Pull request to make Discord integration fully-functional. **Ready for review.**

**Completed**
- Clean up code. Among other streamlining of code and removal of old test code, remove unnecessary functions (`execute_platform_action`,`refresh_access_token` ) and unnecessary model fields (`refresh_token`, `access_token`). Storage of refresh tokens and access tokens can be removed because all calls outside of authentication use `DISCORD_BOT_TOKEN` as the authentication token rather than an access token. Also removed references to `LogAPICall`, which are now unnecessary with fixed action logging system.
- Create a new `DiscordChannel` model to store basic information of Discord channels. This both (1) prevents repeated calls to the Discord gateway for channel information and (2) allows us to match channel IDs and human-readable channel names on PolicyKit. `DiscordChannel`s are populated through `GUILD_CREATE` events received when the websocket connects to the Discord gateway (`handle_guild_create_event` function)
- Using (2) above, allow the user to reference Discord channels to be posted in by name, in addition to channel ID. Makes posting to channels significantly easier for the user, who may not know how to find their Discord channel's ID.
- `DiscordPostMessage` and `DiscordRenameChannel` no longer store `guild_id` (can be retrieved from `DiscordChannel` model). `channel` field renamed to `channel_id` and `id` field renamed to `message_id` to prevent potential bug of ids being non-unique across multiple Discord communities.
- Add `revert` function to `DiscordRenameChannel`. Old channel name is stored as a field of the action (`name_old`), and the Discord gateway can be called to replace the new name with the old name when `revert` is called.
- Replace `is_policykit_action` with new, more accurate check `should_create_action`. Rather than checking messages's author's ids and checking history of `LogAPICall` objects, now checks if messages already have associated `DiscordPostMessage` objects and whether the messages were created within the last two minutes (`2 * settings.CELERY_BEAT_FREQUENCY`). This prevents old messages from being governed by a newly installed PolicyKit.
 - Make `DiscordUser` username unique across communities by setting `username` to `[user_id]:[guild_id]` rather than just `[user_id]`. Prevents `UNIQUE constraint failed` error from occurring if same Discord user tries to install PolicyKit to multiple communities.
 - Allow user to select on a Configure screen which PolicyKit dashboard to sign in to if user belongs to more than one Discord guilds which is integrated with PolicyKit. If user only belongs to one Discord guild which is integrated with PolicyKit, skip the Configure screen and authenticate right away.
 - Immediately check new actions received via websocket against existing policies. Accomplished through manual calls to `consider_proposed_actions` rather than waiting up to a minute for Celery to execute the function.
 - Fix issue where PolicyKit bot was lacking certain permissions not contained in the 'catch-all' Administrator permission
 - Add `PlatformAction`s for following gateway events:
   - Create channel
   - Delete channel
   - Delete message

**Next Steps**
 - Write test suite that tests Discord models and full policies conducted through the Discord integration. Should automatically create test guild and channels within test guild and destroy test guild once testing is concluded.
 - Verify that client receives `heartbeat_ack` between attempts at sending heartbeats. Restart the websocket if `heartbeat_ack` not received.
 - Restart the websocket if the connection is dropped for whatever reason. Handle connection close events appropriately.

### DIFF
--- a/policykit/integrations/discord/apps.py
+++ b/policykit/integrations/discord/apps.py
@@ -1,6 +1,5 @@
 from django.apps import AppConfig
 
-
 class discordIntegrationConfig(AppConfig):
     name = 'integrations.discord'
 
--- a/policykit/integrations/discord/auth_backends.py
+++ b/policykit/integrations/discord/auth_backends.py
@@ -10,20 +10,14 @@ logger = logging.getLogger(__name__)
 
 class DiscordBackend(BaseBackend):
 
-    def authenticate(self, request, oauth=None):
-        if not oauth:
-            return None
-
-        guild_id = request.GET.get('guild_id')
-        logger.info(guild_id)
-
+    def authenticate(self, guild_id, access_token):
         community = DiscordCommunity.objects.filter(team_id=guild_id)
         if not community.exists():
             return None
         community = community[0]
 
         req = urllib.request.Request('https://www.discordapp.com/api/users/@me')
-        req.add_header('Authorization', 'Bearer %s' % oauth['access_token'])
+        req.add_header('Authorization', 'Bearer %s' % access_token)
         req.add_header("User-Agent", "Mozilla/5.0") # yes, this is strange. discord requires it when using urllib for some weird reason
         resp = urllib.request.urlopen(req)
         user_info = json.loads(resp.read().decode('utf-8'))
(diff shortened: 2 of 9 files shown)


amyxzhang:
Nice work!

PR #354, opened 2021-04-29
Commit initial migration files

mashton:
1. Remove `*migrations*` from the gitignore and commit the initial migrations. Going forward we can commit any migration files that get applied on top of these. See [this stack overflow](https://stackoverflow.com/questions/28035119/should-i-be-adding-the-django-migration-files-in-the-gitignore-file#:~:text=Quoting%20from%20the%20Django%20migrations,as%20part%20of%2C%20its%20codebase.&text=If%20you%20follow%20this%20process,conflicts%20in%20the%20migration%20files.) about why migrations should be committed to version control
2. Add instructions in docs about where to find PolicyKit in the browser  (`/main`)
3. Rename `ExternalProcess`->`MetagovProcess`


rickygv99 (approved):

amyxzhang:
Sounds good! Yeah I agree. We originally left it out while the db schema was so in flux but it makes sense now to support migrations.

PR #359, opened 2021-05-05
Add log viewer for policy evaluation logs

mashton:
This PR adds a log viewer, visible at https://policykit.metagov.org/logs. Logged-in users can see all the recent logs for their Community. The celery task cleans up old logs: default behavior is to delete logs that are >1hr old.

The `db_logger` is used to log...
1. Evaluation logs from `execute_policy` function ("X action passed Y policy")
2. Exceptions thrown during policy evaluation, including filter errors
3. `debug` statements added by policy authors

The `db_logger` is _not_ used to log information about actions or anything from integrators. The actstream action log is already sufficient for logging action creation. The purpose of this log viewer, for now anyway, is to debug _evaluation_. So, each log is for a specific `Action+Policy` eval pair.

Policy authors can add `debug` statements like this:
Screen Shot 2021-05-05 at 13 07 51

Which are surfaced in the log viewer like this:
Screen Shot 2021-05-05 at 13 00 52

The database logs are stored in a table for model `EvaluationLog` which is defined in the new  `django_db_logger` app, which is a copy of https://github.com/CiCiUi/django-db-logger customized to fit our needs.


amyxzhang:
Looks great!

Issue #353, opened 2021-04-26
Don't permanently delete removed actions and policies from the database


rickygv99:
Right now, policies which are removed are deleted from the database through calls to Django's `[model].delete()` method. This poses problems for a few reasons:
 1. Policies which are removed contain identifying data which is lost upon deletion. For example, in the action log, policies which are removed are consequently referred to as `[ERROR: platform policy not found]`, since the name is unable to be retrieved.
 2. Policies cannot be recovered. In the event of a mistake or a malicious event, removed policies may be desired to be recovered. However, once removed all data is lost and it is impossible to recover removed policies.
 3. Similar to the above, users may wish to temporarily remove or suspend existing policies. This is impossible in the current framework without direct edits (and subsequent reversions of edits) to policies or manual copying of policies (which may be impossible or infeasible if data is stored in the policies' `datastore` field).

We should add a new Boolean field (`active`) to policies, which is set by default to `True`. Upon removal of a policy, we should set the policy's `active` field to `False` rather than calling `[model].delete()`. `consider_proposed_actions()` should be modified to only call the `filter()` function of active policies. The UI screen for the `RemovePlatformPolicy` and `RemoveConstitutionPolicy` actions should be modified to only list active policies.

We can also add two new `ConstitutionAction`s: `RecoverPlatformPolicy` and `RecoverConstitutionPolicy`. A new UI screen should be created that allows the user to select from inactive policies for recovery.

rickygv99:
Added in #361

PR #361, opened 2021-05-06
Allow user to recover deleted policies and documents

rickygv99:
For #353. Description from issue copied below:

Right now, policies which are removed are deleted from the database through calls to Django's `[model].delete()` method. This poses problems for a few reasons:
 1. Policies which are removed contain identifying data which is lost upon deletion. For example, in the action log, policies which are removed are consequently referred to as `[ERROR: platform policy not found]`, since the name is unable to be retrieved.
 2. Policies cannot be recovered. In the event of a mistake or a malicious event, removed policies may be desired to be recovered. However, once removed all data is lost and it is impossible to recover removed policies.
 3. Similar to the above, users may wish to temporarily remove or suspend existing policies. This is impossible in the current framework without direct edits (and subsequent reversions of edits) to policies or manual copying of policies (which may be impossible or infeasible if data is stored in the policies' `datastore` field).

We should add a new Boolean field (`active`) to policies, which is set by default to `True`. Upon removal of a policy, we should set the policy's `active` field to `False` rather than calling `[model].delete()`. `consider_proposed_actions()` should be modified to only call the `filter()` function of active policies. The UI screen for the `RemovePlatformPolicy` and `RemoveConstitutionPolicy` actions should be modified to only list active policies.

We can also add two new `ConstitutionAction`s: `RecoverPlatformPolicy` and `RecoverConstitutionPolicy`. A new UI screen should be created that allows the user to select from inactive policies for recovery.


policykit/policyengine/migrations/0001_initial.py
```diff
@@ -487,6 +513,17 @@ class Migration(migrations.Migration):
...
+            name='PolicykitRecoverCommunityDoc',
+            fields=[
+                ('constitutionaction_ptr', models.OneToOneField(auto_created=True, on_delete=django.db.models.deletion.CASCADE, parent_link=True, primary_key=True, serialize=False, to='policyengine.ConstitutionAction')),
+                ('doc', models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL, to='policyengine.CommunityDoc')),
+            ],
+            options={
+                'permissions': (('can_execute_policykitrecovercommunitydoc', 'Can execute policykit recover community doc'),),
+            },
+            bases=('policyengine.constitutionaction',),
+        ),
```
mashton, line 526: @rickygv99 we should be adding new migrations instead of updating the existing ones. so, whatever file gets generated from your changes when you run `makemigrations` –– a file prefixed with `0002_`–– should be included in the PR. That way when I or others pull in the changes, we can just apply the new migration file(s). We shouldn't be updating existing migration files, I think, because it will produce errors when trying to `migrate`.

rickygv99: Reverted the changes to `0001_initial.py` and created a new migration file: `0002_policy_recovery.py`

amyxzhang:
cool. new migration file looks good!

PR #364, opened 2021-05-09
Update PolicyKit deployment instructions to work with Metagov

mashton:
When PolicyKit and Metagov are deployed together, some extra steps need to be taken to ensure that the two celery daemons don't conflict. PolicyKit and Metagov each have their own celery and celerybeat daemons the use separate [RabbitMQ virtual hosts](https://www.rabbitmq.com/vhosts.html) as brokers.

I also updated the structure of the Apache2 and Celery setup instructions to be very similar to the metagov equivalents (https://docs.metagov.org/en/latest/installation.html#set-up-celery) so the setup process should be smoother for folks that are installing both at the same time.

This update also includes instructions for creating a dedicated ``celery`` user to run the worker, instead of running with a superuser which [is dangerous](https://docs.celeryproject.org/en/stable/userguide/daemonizing.html#running-the-worker-with-superuser-privileges-root).

### DIFF
--- a/docs/source/gettingstarted.rst
+++ b/docs/source/gettingstarted.rst
@@ -160,37 +160,45 @@ Make sure you have a domain dedicated to Policykit that is pointing to your serv
 
 4. Test your config with ``apache2ctl configtest``. You should get a "Syntax OK" as a response. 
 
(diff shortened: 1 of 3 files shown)


amyxzhang:
Looks good! Thanks for the fixes to parts of the documentation + new documentation!

PR #365, opened 2021-05-09
Use callbacks to get updates from Metagov GovernanceProcesses

mashton:
#### old behavior
PolicyKit would need to fetch the latest `outcome` of an ongoing Metagov process by hitting the `GET /process` endpoint

#### new behavior
PolicyKit receives a callback from metagov each time the `outcome` is changed. The callback is received by the `post_outcome` view in `integrations.metagov.views`. It updates the data in the `MetagovProcess` record.

This is change was made possible by https://github.com/metagov/metagov-prototype/pull/42. More context in https://github.com/metagov/metagov-prototype/issues/19.

### DIFF
--- a/policykit/policykit/settings.py
+++ b/policykit/policykit/settings.py
@@ -239,6 +239,7 @@ LOGGING = {
 
 from celery.schedules import crontab
 
+# Replace with "amqp://USERNAME:PASSWORD@localhost:5672/VIRTUALHOST"
 CELERY_BROKER_URL = 'amqp://'
 CELERY_RESULT_BACKEND = 'django-db'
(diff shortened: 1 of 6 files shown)


policykit/policykit/settings.py
```diff
@@ -239,6 +239,7 @@
 
 from celery.schedules import crontab
 
+# Replace with "amqp://USERNAME:PASSWORD@localhost:5672/VIRTUALHOST"
```
mashton, line 242: This belongs with  https://github.com/amyxzhang/policykit/pull/364 - now that both Metagov and PolicyKit are using celery, we need to use a RabbitMQ virtual host as the broker

PR #372, opened 2021-05-11
Add jsonform editor for Metagov plugin configuration

mashton:
Use [jsonform](https://github.com/jsonform/jsonform) to generate a form for the Metagov config section under PolicyKit Settings page. It also surfaces the webhook URLs to the admin (since they need to go register them with each platform).

The diff is big because it includes `jsonform` and `underscore` js packages.

Screen Shot 2021-05-11 at 18 36 32

**This branch contains https://github.com/amyxzhang/policykit/pull/371 too, that one should be merged first


amyxzhang:
Nice! Just flagging that in the future, we will want more descriptions with these forms.

Issue #305, opened 2021-03-17
Proposal status is coupled to "action.execute()"


mashton:
This issue concerns non-bundled PlatformActions. I'm not sure if the same issue exists for other kind of actions.

**Problem**
We only mark a PlatformAction's proposal as `PASSED` when `action.execute()` is invoked ([here](https://github.com/amyxzhang/policykit/blob/4135083a879a1621ccfe5856db38da1033bdd515/policykit/policyengine/models.py#L773)). This assumes that the policy author calls `action.execute()` in the `pass_policy` block, which is not always the case. If the author does NOT call `action.execute()` in the `pass_policy` block, then the Proposal is stuck in `proposed` state forever.

**Solution**
We should decouple the proposal status from action execution and reversion. We can do this by immediately setting `Proposal.status` to `passed` when `pass_policy` is run, and `failed` when `fail_policy` is run. & removing [this line](https://github.com/amyxzhang/policykit/blob/4135083a879a1621ccfe5856db38da1033bdd515/policykit/policyengine/models.py#L773) which marks an action as `passed` when it is executed. With this, execution will be decoupled from the proposal status, and the proposal will be correctly marked "passed" when it passes, regardless of whether the policy author called execute() in the pass block or not.

mashton:
Partially fixed with commit https://github.com/amyxzhang/policykit/pull/300/commits/29506626d94db1b70ce8b5f31fc704046b94f8d9, but we should stop calling "pass_action" from within "execute." Leaving this issue open for that work.

mashton:
closed by https://github.com/amyxzhang/policykit/pull/387 and https://github.com/amyxzhang/policykit/pull/389

PR #387, opened 2021-05-18
Remove remaining calls to pass_action from execute functions

rickygv99:
Remove the remaining calls to `pass_action()` from `execute()` functions in various models in `policyengine/models.py`, `discord/models.py` and `discourse/models.py`. Since `pass_action()` is already called in `_execute_policy()`, `pass_action()` was being called twice on all actions. Since `action.send()` sends an `action was passed` log when executed and since it is called whenever `pass_action()` is called, the log was sending twice, thus appearing that all logs were duplicated in the homepage logging.

### DIFF
--- a/policykit/integrations/discord/models.py
+++ b/policykit/integrations/discord/models.py
@@ -127,8 +127,6 @@ class DiscordPostMessage(PlatformAction):
             self.community_post = self.message_id
             self.save()
 
-        super().pass_action()
-
 class DiscordDeleteMessage(PlatformAction):
     channel_id = models.IntegerField()
     message_id = models.IntegerField()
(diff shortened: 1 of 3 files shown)


mashton:
@rickygv99 thanks for starting this. I had left https://github.com/amyxzhang/policykit/issues/305 open for doing exactly this.

There are some more changes needed for proposal status to be completely decoupled from `execute`, though. As-is this change introduced some new bugs. One issue is that there are lots of places in the codebase with logic like this:

```python
#if they have execute permission, skip all policies
if action.initiator.has_perm(action.app_name + '.can_execute_' + action.action_codename):
     action.execute()
else:
     # try execute_policy for each policy
```

Now, all the actions that are executed because of `.can_execute_` permissions will remain in with `proposed` status indefinitely, because pass_action is never called. There might be other places that we expect `execute` to `pass` an action, too. I can go fix the aforementioned issue with `can_execute_` right now, and look for other instances.

And sorry about the duplicate logging –– I didn't realize there was going to be an issue with calling `pass_action` twice. We could call `action.send(action, verb='was passed')` from `execute_policy` instead –– that way it only gets logged once, even if `pass_action` is called repeatedly (which of course, it shouldn't be).

amyxzhang:
Sorry, should I revert the merge?

mashton:
No it's OK, it's not deployed anywhere. I'm opening a PR now

PR #389, opened 2021-05-18
Call pass_action when the engine executes an action

mashton:
Follow-up to https://github.com/amyxzhang/policykit/pull/387 to fix a bug. The bug was that actions executed because the user had `can_execute_*` permission were left in "proposed" state. Additionally some of the `has_perm` checks for can_execute were not working properly because they used the wrong app label prefix.

The code be easiest to review commit-by-commit, since I included a refactor to pull out some repeated code.
This commit is what fixes the bug: https://github.com/amyxzhang/policykit/commit/10e85a07293096f9a8c24c16d5e1969751c9975c

### DIFF
--- a/policykit/tests/test_policy_evaluation.py
+++ b/policykit/tests/test_policy_evaluation.py
@@ -5,7 +5,7 @@ from django.test import Client, TestCase
 from integrations.metagov.library import update_metagov_community, metagov_slug
 from integrations.metagov.models import MetagovProcess, MetagovPlatformAction
 from integrations.slack.models import SlackCommunity, SlackPinMessage, SlackUser
-from policyengine.models import CommunityRole, PlatformPolicy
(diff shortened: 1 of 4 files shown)


policykit/tests/test_policy_evaluation.py
```diff
@@ -46,6 +43,70 @@ def setUp(self):
...
+
+        # action initiated by user with "can_execute" should pass
+        action = PolicykitAddCommunityDoc(name="my doc", initiator=user_with_can_execute, community=self.community)
+        action.save()
+        self.assertEqual(action.proposal.status, "passed")
+
+        # action initiated by user without "can_execute" should fail
+        action = PolicykitAddCommunityDoc(name="my other doc", initiator=self.user, community=self.community)
+        action.save()
+        self.assertEqual(action.proposal.status, "failed")
```
mashton, line 108: This test demonstrates the fixed bug

rickygv99 (approved):

PR #390, opened 2021-05-19
Add ability to upload and download policies

rickygv99:
Previously, it was tedious to try out policies that someone else developed or to recreate policies on new communities. Each code block had to be copy/pasted individually. This PR allows users to download policies (from both the editor and from the list of passed policies) into a text file (containing a JSON representation of the policy). Users are also able to upload policies -- a new option is added to the Platform Policies and Constitution Policies menus on `actions.html` called `Upload Platform Policy` and `Upload Constitution Policy` respectively. Upon clicking either option, a file dialog is opened. The user can select a policy from the file dialog and the editor will be opened with the policy automatically loaded in.

Screen Shot 2021-05-19 at 2 53 35 PM
Screen Shot 2021-05-19 at 2 53 44 PM
Screen Shot 2021-05-19 at 2 54 03 PM


amyxzhang:
I prefer not to have the upload option in the set of possible actions. Instead I'd want it on the editing page.

PR #392, opened 2021-05-23
Correction for DRIVER_EVENT_RECEIVER_URL place

hozzjss:
DRIVER_EVENT_RECEIVER_URL is an env that only exists within metagov, and is only used by metagov, and also pk doesn't use .env files anywhere


amyxzhang:
Thank you!

PR #395, opened 2021-05-26
Add Metagov example for Discourse web monetization

mashton:


docs/source/sample_policies.rst
```diff
@@ -467,3 +467,156 @@ This policy can be defined for any PolicyKit community (a Slack community, for e
...
+    custom_wallet_field_key = "1"
+    old_wallet = action.event_data.get("old_user_fields", {}).get(custom_wallet_field_key)
+    new_wallet = action.event_data.get("user_fields", {}).get(custom_wallet_field_key)
+    if old_wallet == new_wallet:
+      debug(f"no wallet change for {user}, they must have changed another field. skipping.")
+      return False
+
+    debug(f"User {user} changed their wallet from '{old_wallet}' to '{new_wallet}'")
+    action.data.set("old_wallet", old_wallet)
+    action.data.set("new_wallet", new_wallet)
```
mashton, line 498: @amyxzhang I follow this pattern frequently, using `action.data` to store variables that I need to access in future evaluation steps. This might become an issue when actions are evaluated against multiple policies, because there would be other policies that use the same key in the datastore. Really I only need these variables to be local to the action+policy pair, so maybe we could introduce a datastore just for that. Super low importance, I think, but just a thought

PR #399, opened 2021-05-31
Constitution action for roles votes keep notifying

hozzjss:
### Test plan
- Try starting a vote over a role operation
- If the notify action works once then great


mashton:
This *might* be a completely unrelated bug, but might not: I opened a fix for a ConstitutionAction bug here https://github.com/amyxzhang/policykit/pull/402. New ConstitutionActions were not being saved in a specific case. This could lead to the behavior you are describing- Constitution Actions that continuously notify- since they were not appropriately being failed when they should have been.

policykit/policyengine/views.py
```diff
@@ -646,10 +646,10 @@ def role_action_users(request):
     action.community = user.community
     action.initiator = user
     action.role = CommunityRole.objects.filter(name=data['role'])[0]
-    action.save()
+    action.save(without_govern=True)
     action.users.set(CommunityUser.objects.filter(username=data['user']))
     action.ready = True
-    action.save()
+    action.save(treat_as_new=True)
```
mashton, line 652: @amyxzhang @rickygv99 what's the desired behavior here? I'm not sure what should happen.

It looks like there is already a special case for `PolicykitAddUserRole` and `PolicykitRemoveUserRole` which override the "shouldCreate" function to use this `ready` flag, as a way to defer policy evaluation until the second time the model is saved... So that seems similar to what @hozzjss is trying to introduce here with "treat_as_new"

https://github.com/amyxzhang/policykit/blob/da19825490164e770b83c9339a3775ae02bd28d9/policykit/policyengine/models.py#L751-L752

policykit/policyengine/models.py
```diff
@@ -512,10 +512,16 @@ def shouldCreate(self):
...
     def save(self, *args, **kwargs):
+        kwargs_clean = kwargs.copy()
+        for key in 'treat_as_new', 'without_govern':
+            if kwargs.get(key):
+                kwargs_clean.pop(key)
+
         """
         Saves the action. If new, checks it against current passed policies. Note: Only meant for internal use.
         """
-        if self.shouldCreate():
```
mashton, line 518: I'm guessing the point of this was so that subclasses could override it like this
https://github.com/amyxzhang/policykit/blob/da19825490164e770b83c9339a3775ae02bd28d9/policykit/policyengine/models.py#L751-L752
@amyxzhang or @rickygv99 would know more

hozzjss:
Now i understand

hozzjss:
I’ll keep this pr as a place for discussion i think .ready is much better than my solution i should have just set that instead of this whole work around

hozzjss:
I created a different way of doing things through making just like pk is a one time thing ready also should be a one time thing through a hack @mashton

policykit/policyengine/models.py
rickygv99: Yes, for deferring policy evaluation, the `shouldCreate` function should be overridden.

PR #400, opened 2021-05-31
Fix voting to handle Constitution Actions

hozzjss:
Following fixes are applied:

Added placeholder reaction on voting message just to be cool
fix policy breaking when checking for a maybe value 
Policy breaks when checking for community_origin a field that's optional 
used `hasattr` instead
update discord democracy starterkit to work
Accept votes on constitution policies in discord


amyxzhang:
This all looks good to me, thanks. I love the idea of reaction placeholders!

PR #405, opened 2021-06-04
Add link to install and authorize Metagov Slack Plugin on Settings page

mashton:
This demonstrates how to activate the Metagov Slack Plugin using an authorization flow that is initiated by Metagov.
Uses new Metagov endpoints from https://github.com/metagov/metagov-prototype/pull/69


mashton:
Screen Shot 2021-06-04 at 12 17 48

@amyxzhang this PR just adds a link to initiated the Slack Authorization flow to enable the Metagov Slack plugin (similar to "Add to Slack" button for installation). I don't know if we need to merge this in, it is more of a proof of concept. The real desired behavior is for the "Add To Slack" button on `policykit.org/login` to use the new [Metagov /authorize endpoint](https://prototype.metagov.org/redoc/#operation/Authorize%20plugin) to install the Slack app. I'll get started on that next.

amyxzhang:
Thanks! I agree we don't need to merge this in - ideally, the end user shouldn't need to know about MetaGov, right? Are there points where you think they'll have to?

mashton:
I don't think the user needs to know about metagov, no.

PR #411, opened 2021-06-25
Use Metagov slack.emoji-vote for boolean PlatformAction votes and elections

mashton:
Use the [slack.emoji-vote](https://prototype.metagov.org/redoc/#operation/Start%20slack.emoji-vote) metagov governance process to administer two kinds of votes: (1) Boolean votes on PlatformActions, and (2) Choice votes on PlatformActionBundle elections. Slack election voting was broken before, this fixes it.

Screen Shot 2021-06-25 at 15 30 55
Screen Shot 2021-06-25 at 15 30 48

Follow-ups:
- Support ConstitutionActions https://github.com/amyxzhang/policykit/issues/412
- Handle closing processes so we don't have dangling pending processes in metagov

### DIFF
--- a/policykit/integrations/slack/models.py
+++ b/policykit/integrations/slack/models.py
@@ -94,7 +94,8 @@ class SlackCommunity(CommunityPlatform):
             headers={"X-Metagov-Community": self.metagov_slug},
         )
         if not response.ok:
-            raise Exception(f"Error: {response.status_code} {response.reason} {response.text}")
+            logger.error(f"Error making Slack request {method_name} with params {values}")
+            raise Exception(f"{response.status_code} {response.reason} {response.text}")
         if response.content:
             return response.json()
         return None
@@ -165,9 +166,6 @@ class SlackCommunity(CommunityPlatform):
             logger.debug("Ignoring bot event")
             return
 
-        event_type = outer_event["event_type"]
-        initiator = outer_event["initiator"].get("user_id")
-
         from integrations.slack.views import maybe_create_new_api_action
 
         new_api_action = maybe_create_new_api_action(self, outer_event)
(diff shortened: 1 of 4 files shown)


policykit/integrations/slack/views.py
```diff
@@ -235,92 +220,74 @@ def maybe_create_new_api_action(community, outer_event):
...
-            else:
-                a = action.bundled_actions.all()[0]
-                values["channel"] = a.channel
+    # Kick off process in Metagov
+    logger.debug(f"Starting slack vote on {action} governed by {policy}. Payload: {payload}")
+    response = requests.post(
+        f"{settings.METAGOV_URL}/api/internal/process/slack.emoji-vote",
+        json=payload,
+        headers={"X-Metagov-Community": policy.community.metagov_slug},
+    )
```
mashton, line 274: This is where the vote gets kicked off, in the `post_policy` function.

policykit/integrations/slack/models.py
```diff
@@ -176,41 +174,107 @@ def handle_metagov_event(self, outer_event):
...
+                        )
+                        NumberVote.objects.create(proposal=voted_action.proposal, user=user, number_value=num)
+                    elif existing_vote.number_value != num:
+                        logger.debug(
+                            f"Casting number vote {num} by {user} for {voted_action} in bundle {action_bundle} (vote changed)"
+                        )
+                        existing_vote.number_value = num
+                        existing_vote.save()
+
+    def post_message(self, text, users=[], post_type="channel", channel=None):
```
mashton, line 229: This function doesn't get called by anything, but, I created it just because I didn't want to throw away this code. The `im` and `ephemeral` post types didn't make sense to keep in `notify_action`, because you can't perform votes in those contexts. As discussed, we should separate out the authoring API "voting" from "notification," at which point we can re-use this code for notifying users.

policykit/integrations/slack/models.py
```diff
@@ -176,41 +174,107 @@ def handle_metagov_event(self, outer_event):
...
-                    NumberVote.objects.create(proposal=voted_action.proposal, user=user, number_value=num)
+            logger.debug(f"PlatformAction saved: {new_api_action.pk}")
+
+    def handle_metagov_process(self, process):
+        """
+        Handle a change to an ongoing Metagov slack.emoji-vote GovernanceProcess.
+        This function gets called any time a slack.emoji-vote associated with
+        this SlackCommunity gets updated (e.g. if a vote was cast).
+        """
+        assert process["name"] == "slack.emoji-vote"
```
mashton, line 185: This is where the SlackCommunity tallies up votes and creates BooleanVote and NumberVote models for each. This gets called each time the governance process is updated. The process object looks like this:

```
{
  "id": 18,
  "name": "slack.vote",
  "community": "30ab507c-5f88-4194-95a0-4f583dd9a400",
  "status": "completed",
  "errors": {},
  "outcome": {
    "url": "https://testworkspace-i0j9034.slack.com/archives/C01G68P700Z/p1624566936002800",
    "channel": "C01G68P700Z",
    "message_ts": "1624566936.002800",
    "votes": {
      "yes": {
        "users": [
          "U025LEMKQP7",
           "U025LEMKQ88"
        ],
        "count": 2
      },
      "no": {
        "users": [
        ],
        "count": 0
      }
    }
  }
}
```

policykit/integrations/slack/models.py
mashton: https://github.com/amyxzhang/policykit/issues/413

amyxzhang: Confused - why can't you perform votes in those contexts? I originally had those in (though not really fleshed out) because you might want private voting.

Clarifying my thoughts below - does this line up with what you are thinking?

Yes we should separate the act of sending a message to users in some way - which we want policy authors to be able to do whenever. You should be able to send any kind of message in private, public, mpim, ephemeral, etc. This should be like a "post_message()" function.

In the case of voting, there should be a function for enacting a voting, like "call_vote()", which then itself calls post_message() so that people have something to vote on. So settings for how to "notify" people about a vote need to be passed by the policy author through the voting function.

amyxzhang:
Looks good!

policykit/integrations/slack/models.py
mashton: > Confused - why can't you perform votes in those contexts? I originally had those in (though not really fleshed out) because you might want private voting.

You can't perform votes on `ephemeral` message because they don't support emoji-reactions. `im` voting makes sense, actually, you're right. It just wasn't implemented yet (there was/is no support for tallying votes on multiple `community_post`s). Agreed this should be a feature to add (to the Metagov Slack emoji-vote process)!

> Yes we should separate the act of sending a message to users in some way - which we want policy authors to be able to do whenever. You should be able to send any kind of message in private, public, mpim, ephemeral, etc. This should be like a "post_message()" function.

yes agreed

> In the case of voting, there should be a function for enacting a voting, like "call_vote()", which then itself calls post_message() so that people have something to vote on. So settings for how to "notify" people about a vote need to be passed by the policy author through the voting function.

Does it *need* to work like this internally? We are using Metagov, so, the fact that the voting mechanism uses the same Slack request as "post_message" internally is something that is hidden in the Metagov Slack plugin (and yes, [it currently does](https://github.com/metagov/metagov-prototype/blob/6d4506ec9ada78e3d538280d5f56d53bd7a077fe/metagov/metagov/plugins/slack/models.py#L210)).

The way I implemented, it will work more like this:
* VOTING: Policy author writes `slack_community.perform_vote(kind="mpim", users=["U1","U2","U3"])`. The `perform_vote` function makes a request to Metagov: `POST /api/internal/process/slack.emoji-vote {"message_type": "mpim", "users": [..], "text: "..." }`
* MESSAGING: Policy author writes `slack_community.post_message(kind="mpim", users=["U1","U2","U3"])`. The `post_message` function makes a request to Metagov: `POST /api/internal/action/slack.post-message {"kind": "mpim", "users": [..], "text": "..." }`

amyxzhang: Thanks for the pointer regarding ephemeral! I don't think we need it for voting then since you can provide that same functionality by letting a policy delete a message after some time/condition. I also think this is not high priority.

Well I don't have a strong opinion about WHERE it happens, in PK or MG, but I just wanted to point out that one will need to post a message when eliciting votes since emoji votes are only on messages (and right now in PK, we have to store which message ID so we know where to check for emoji votes). If you've got it covered in MG that sounds good.

amyxzhang: what about "initiate_vote()" or "initiate_vote_process()" or "initiate_voting()" instead of "perform_vote()"? Thinking about whether it could get confused with making an individual vote or carrying out a vote.

mashton: 👍 I like the "initiate_" options you mentioned. I won't make any changes to the authoring API without discussing with you & documenting them.

PR #419, opened 2021-06-30
Fixes to Editor

rickygv99:
Updates / fixes to editor so it works properly
 - Integrated PyLint into the editor. Now any Python syntax errors will be caught and displayed by the editor.
 - Now the editor is able to detect more than one error at a time. Previously, the editor navigated the AST tree of the code and stopped at the first error it saw (it couldn't navigate past an error). All errors are now shown.
 - Fixed tooltips not displaying. Now when the user mouses over a gutter marker (on lines with errors), a tooltip will appear with the line number, character number, and error message pertaining to the error.
 - Fixed Codemirror CSS / JavaScript plugins not loading in properly.

Note: This doesn't pertain to the filter, just the editing experience. Dealing with intermediate variables being used to bypass the filter still needs to be handled.


amyxzhang:
Yay exciting!!

PR #422, opened 2021-07-01
Skip policies that raise exceptions

mashton:
The policy author can review the error on the Logs page. This protects against situations where a policy consistently raises an exception, so the PlatformAction gets stuck in `proposed` state and we just try again and again to execute the same failing policy.


amyxzhang:
Is the merge conflict due to @rickygv99's reverted PR?

mashton:
Yep it was just a conflict with some new imports, just updated this branch

Issue #346, opened 2021-04-16
Implement proposing platform actions in the UI


mashton:
These buttons don't do anything:

<img width="580" alt="Screen Shot 2021-04-16 at 17 10 38" src="https://user-images.githubusercontent.com/5333982/115084306-af874580-9ed6-11eb-9c91-6ee63f02e1b4.png">

use model admin https://docs.djangoproject.com/en/3.2/ref/contrib/admin/ if possible, else forms https://docs.djangoproject.com/en/3.2/topics/forms/

Issue #318, opened 2021-04-02
Lift "execute" and "revert" into policy


mashton:
### Problem: the code assumes that policy author wants to "govern" the trigger action

Currently PolicyKit reverts the "trigger" action if `check` returns `None` (which means “proposed”) or `FAILED` . It [happens here in the engine](https://github.com/amyxzhang/policykit/blob/41c620d8b473fb40cc0f60293891979f60353f14/policykit/policyengine/models.py#L831-L838) and [here in the slack integration](https://github.com/amyxzhang/policykit/blob/f5763ba756cfebed08c1e2e9880b2e75581c058c/policykit/integrations/slack/views.py#L266) and similarly in other integrations. (Related: dedup the policy evaluation https://github.com/amyxzhang/policykit/pull/300)

### Short-term solution (current)
If we want to write a different kind of policy that _doesn't_ "govern" the trigger action, we can just write the policy code into the `pass` block.

This works for some cases, but has two issues:
1) It is confusing for the policy author. The block names are no longer logical
2) We cannot use PolicyKit's polling features (continually re-evaluating proposed policies using celery beat), because that relies on special functionality of the `check` block.

### Proposed solution
Add another policy block to let the _author_ decide what to do when `check` returns `None`(proposed) or `FAILED`. Here’s an example of what that might look like to the author:

```
Check:
    if something:
        return PENDING # I'm waiting for a decision, please run the pending block
    elif something_else:
        return PASSED # the check passed, please run the pass block
    elif another_thing:
        return FAILED # the check passed, please run the fail block
    return None #error, check must return a valid state
Pass:
	action.execute()
Pending:
	action.revert()
Fail:
	action.revert() # if check has already returned PENDING previously, then this does nothing because the action would have already been reverted
```

#### Related issues
* This solution will be easier to implement once the policy evaluation refactor https://github.com/amyxzhang/policykit/pull/300 is done
* This is related to issue https://github.com/amyxzhang/policykit/issues/305, which surfaces the fact that the `execute()` is tightly coupled to `PASSED` state. I think https://github.com/amyxzhang/policykit/issues/305 should be fixed first. In order to express more kinds of policies, we want to decouple `revert` and `execute` from `passed` and `failed` states respectively.
   * (In the future, we could have an abstraction on top of the base PlatformPolicy for "policy-that-governs-an-action" which implements the default `pass`/`pending`/`fail` blocks above, to make it easier for the policy author).

amyxzhang:
[...]

PR #429, opened 2021-07-12
Allow proposing any PlatformAction in the PolicyKit UI

mashton:
* Implements the UI for proposing PlatformActions (Closes https://github.com/amyxzhang/policykit/issues/346)
* Changes the PolicyEngine evaluation loop so that....
  * When an action that was proposed in the UI passes (aka `community_origin == False`), the engine calls `action.execute()`.
     * Another way I could have implemented this is to modify the "pass" of All Actions Pass policy to be `if not action.community_origin: action.execute()`. That would also work and aligns with the goals of https://github.com/amyxzhang/policykit/issues/318, but I'd rather not mess with the StarterKit policies now and just continue as-is adding this logic to the engine. Then when we implement 318 we can lift this logic into policies (if we agree on that - want to confirm approach with @amyxzhang first as well). I think this approach will be less disruptive to current users.
  * When a constitution action passes, the engine executes it (@rickygv99 this removes the need for "all actions execute" fallback policy, which was removed. you may need to change that starter policy on your server.)

-----
screenshots

### All the available actions for the current platform are listed. The "action_codename" is used for now because we don't have readable names defined.
Screen Shot 2021-07-12 at 12 09 59

### The form is generated based on the model's fields. The choice of fields can be overridden on each model. For example, here we hide the "previous name" field because it's not necessary:
Screen Shot 2021-07-12 at 12 10 34

### If there are no overrides, all fields are shown (Excluding BaseAction and PlatformAction fields). For example:

Screen Shot 2021-07-12 at 12 15 30

### On submit, a message shows whether the action passed, failed, or was proposed.
 
Screen Shot 2021-07-12 at 12 12 10
Screen Shot 2021-07-12 at 12 12 29

### DIFF
--- a/policykit/policyengine/views.py
+++ b/policykit/policyengine/views.py
@@ -5,12 +5,14 @@ from django.contrib.auth.models import Permission
 from django.http import HttpResponse, JsonResponse, HttpResponseBadRequest, HttpResponseRedirect
 from django.views.decorators.csrf import csrf_exempt
 from django.shortcuts import render, redirect
+from django.forms import modelform_factory
+from django.apps import apps
 from actstream.models import Action
-from policyengine.filter import *
+from policyengine.filter import filter_code
 from policyengine.linter import _error_check
+from policyengine.utils import find_action_cls, get_action_codenames
 from policykit.settings import SERVER_URL, METAGOV_ENABLED
 import integrations.metagov.api as MetagovAPI
-import urllib.parse
 from urllib.parse import quote
 import random
 import logging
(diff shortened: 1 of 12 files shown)


policykit/policyengine/views.py
```diff
@@ -814,12 +856,20 @@ def _execute_policy(policy, action, is_first_evaluation: bool):
...
         assert action.proposal.status == Proposal.PASSED
 
+        # EXECUTE the action if....
+        # it is a PlatformAction that was proposed in the PolicyKit UI
+        if issubclass(type(action), PlatformAction) and not action.community_origin:
+            action.execute()
+        # it is a constitution action
+        elif issubclass(type(action), ConstitutionAction):
+            action.execute()
+
```
mashton, line 866: New logic in `pass` block.

policykit/policyengine/views.py
```diff
@@ -832,20 +882,20 @@ def _execute_policy(policy, action, is_first_evaluation: bool):
...
+        action.community_origin
 
-    elif check_result == Proposal.PROPOSED and is_first_evaluation:
-        # Revert if this action originated in the community (ie it was not proposed manually in the PK UI)
-        if hasattr(action, 'community_origin'):
-            debug(f"Reverting")
-            action.revert()
+    if should_revert:
+        debug(f"Reverting")
+        action.revert()
```
mashton, line 893: This logic didn't change, I just tried to make it more readable.

amyxzhang:
Looks great, thank you for getting this done!

I agree with the decision to hold off on policy structuring changes, though we should do that soon anyway (hopefully).

PR #430, opened 2021-07-13
Fix database constraint issue in Discord

mashton:
[Discord snowflakes](https://discord.com/developers/docs/reference#snowflakes) are 64 bit integers, so we need to use BigIntegerField


policykit/integrations/discord/models.py
```diff
@@ -128,11 +127,10 @@ def execute(self):
             self.save()
 
 class DiscordDeleteMessage(PlatformAction):
-    channel_id = models.IntegerField()
-    message_id = models.IntegerField()
+    channel_id = models.BigIntegerField()
+    message_id = models.BigIntegerField()
     text = models.TextField(blank=True, default='')
 
-    ACTION = f"channels/{channel_id}/messages/{message_id}"
```
mashton, line 135: @rickygv99 I'm deleting these because they don't make sense / aren't used. Tested it out in the django shell as I was trying to understand it:

```
In [1]: d = DiscordDeleteMessage(channel_id=123, message_id=123, text="Asdfasd")

In [2]: d.ACTION
Out[2]: 'channels/<django.db.models.fields.IntegerField>/messages/<django.db.models.fields.IntegerField>'
```

rickygv99: Okay, if they aren't used anywhere, let's delete them in that case. Might be unused legacy code.

rickygv99 (approved):

PR #435, opened 2021-07-13
Fix starter-kits [WIP]

rickygv99:
**_Work in progress._**

- [x] Starter-kit initialization script `scripts/starterkits.py` is over 1,400 lines of code right now. Much of the code is recreation of starter-kit code for different platforms. Refactor starter-kit initialization script to remove code duplication.
- [X] Don't hard-code starter-kit information into starter-kit initialization code. Instead, store starter-kit information in JSON text files and read information in using a new function `init_starter_kit(filename)`.
- [x] Move all platform starter kit initialization code out of platform-specific `models.py` files. The initialization code is largely the same across all platforms. Move into a single platform-agnostic function in `CommunityPlatform` in `policyengine/models.py` so we don't have to recreate across every platform.
- [x] Permissions should be defined once for each platform. No separate `VIEW`, `ADD`, and `EXECUTE` permissions lists. Can we combine this with the `ACTIONS` list as well? Would be cleaner that way.
- [x] Remove use of `StarterKit`, `GenericPolicy` and `GenericRole` models.
- [ ] Fix all starter-kits so that they function correctly.
- [ ] Add tests for all starter-kits.
- [ ] Create a new, flexible starter-kit UI which can accommodate different amounts of starter-kits. Maybe a scrollview list?
- [x] Move starter-kit names and descriptions from being hardcoded into UI and into JSON files, so all starter-kit info is in one place.
- [x] Starter-kits should initialize automatically when PolicyKit is initialized. There should not be a separate script since (a) the script is only called once, upon initialization, and (b) PolicyKit can't be initialized for a community until this script is run.


amyxzhang:
BTW I think @mashton is doing some work to standardize the policykit API across the different platforms.

amyxzhang:
I will say that I quite dislike the starterkit UI right now so please feel free to completely redo :p 

We might also think about a nicer way to store different starterkits in our code repository for the future when others may want to contribute starterkits to our code base. (Separately, eventually we'll also want a way for people to publish their policies or policy snippets to a shared space on policykit.org but those don't need to be in every install's "starterkit" set.)

policykit/starterkits/dictator.txt
```diff
@@ -0,0 +1,124 @@
...
+        "Can delete boolean vote",
+        "Can view boolean vote",
+        "Can add number vote",
+        "Can change number vote",
+        "Can delete number vote",
+        "Can view number vote",
+        "Can add platformactionbundle",
+        "Can add platformpolicybundle",
+        "Can add constitutionactionbundle",
+        "Can add constitutionpolicybundle",
```
mashton, line 86: These will need to be updated to remove nonexistent permissions in the same way that https://github.com/amyxzhang/policykit/pull/448 did

policykit/policyengine/models.py
```diff
@@ -335,34 +324,6 @@ def make_api_call(cls, community, values, call, action=None, method=None):
         res = community.make_call(call, values=values, action=action, method=method)
         return res
 
-class GenericPolicy(models.Model):
```
mashton, line 338: These models are removed, so there should be migration files for deleting those tables from the database. Can you run `makemigrations` and commit the result?

policykit/policyengine/utils.py
```diff
@@ -46,4 +48,18 @@ def construct_authorize_install_url(request, integration, community=None):
...
\ No newline at end of file
+    return url
+
+def get_starterkits_info():
+    """
+    Get a list of all starter-kit names and descriptions.
+    """
+    starterkits = []
+    for kit_file in os.listdir('./starterkits'):
+        f = open(f'{os.getcwd()}/starterkits/{kit_file}')
```
mashton, line 59: Has this been tested on a server that's serving up files using apache web server? Just want to check, since the `cwd` command there might differ from what you get locally.

policykit/policyengine/views.py
```diff
@@ -611,25 +612,96 @@ def clean_up_proposals(action, executed):
...
+    from policyengine.models import Proposal, CommunityPlatform, Policy, CommunityRole, CommunityUser
+
+    post_data = json.loads(request.body)
+
+    logger.debug(f'Initializing with starter kit: {post_data["starterkit"]}')
+    f = open(f'{os.getcwd()}/starterkits/{post_data["starterkit"]}.txt')
+
+    kit_data = json.loads(f.read())
+
+    community = CommunityPlatform.objects.get(community_name=post_data["community_name"])
```
mashton, line 628: The `community_name` is not necessarily unique, we should probably use primary key instead. No need to change it in this PR, though, since that was already an issue.

rickygv99: Ah, true! Left a comment there

policykit/policyengine/views.py
```diff
@@ -611,25 +612,96 @@ def clean_up_proposals(action, executed):
...
-    starter_kit = StarterKit.objects.get(name=starterkit_name, platform=platform)
-
-    community = CommunityPlatform.objects.get(community_name=community_name)
-    starter_kit.init_kit(community, creator_token)
-
-    logger.info('starterkit initialized')
-
-    redirect_route = request.GET.get("redirect")
-    if redirect_route:
-        return redirect(f"{redirect_route}?success=true")
```
mashton, line 630: This `redirect` parameter was used when a user installs a second platforms from the "settings" page. For example if I'm logged in with Discord, and I want to install to my Slack too, the flow looks like this:
1. Navigate to /main/settings
2. On the top right, click Add Integration > Slack
3. Approve Slack consent screen
4. Now I see the starterkit page and select a starterkit to use with my new SlackCommunity. Click on a starterkit.
5. I am redirected back to /main/settings

That flow is broken by this change.  Ideally we can fix that before merging it in @rickygv99 @amyxzhang

policykit/policyengine/utils.py
rickygv99: Tested and updated!

policykit/policyengine/views.py
rickygv99: Updated!

mashton (commented):
Looks good to me once the tests are fixed

policykit/policyengine/views.py
mashton: great nice work!!

amyxzhang:
Ok going to merge this!

Issue #413, opened 2021-06-25
Split `notify_action` into 2 or more functions (performing a vote vs. notifying users)


mashton:
Currently the function `notify_action` is used to start a vote. It starts a choice vote if the action is of type ELECTION, otherwise it starts a boolean vote. This is the function signature for each platform community subclass:

```
# SlackCommunity
def notify_action(self, action, policy, users=None, post_type="channel", template=None, channel=None)
# DiscordCommunity
def notify_action(self, action, policy, users=None, template=None, channel=None)
# DiscourseCommunity
def notify_action(self, action, policy, users=None, template=None, topic_id=None)
# RedditCommunity
def notify_action(self, action, policy, users=None)
```

#### Proposed change

* Rename `notify_action` to `vote_on_action`, keep same parameters
* Add function `post_message` to PlatformCommunity subclasses with this signature:

      ```py
      # Slack, Discord
      def post_message(self, text, users=[], post_type="channel", channel=None)
      # Discourse
      def post_message(self, text, users=[], topic_id=None)
      # Reddit
      def post_message(self, text, users=[])
      ```

mashton:
@amyxzhang what function names do you want for voting and messaging? I think you made a suggestion somewhere in slack but I can't find it now

mashton:
Note to self:  update the function signature to make it possible for `post_message` to post a message in response to an existing thread on Slack.

amyxzhang:
I forget where I put it. I think "vote_on_action" might be misinterpreted to mean making a vote so I think I proposed "initiate_vote" instead. 

And "post_message" sounds good.

PR #408, opened 2021-06-16
Make Slack integration use Metagov, rename `PlatformCommunity`, reset migrations

mashton:
* Change the Slack integrator to use Metagov instead of connecting directly to Slack. (Closes https://github.com/amyxzhang/policykit/issues/406). Authorization is handled by Metagov.
* Data model change for supporting multi-platform communities it related to https://github.com/amyxzhang/policykit/issues/384. This is the new data model:
```py
Community(Model):  # NEW "parent community"
- id: AutoField(unique) # primary key
- metagov_slug: CharField(unique)  # identifies this Community in Metagov

PlatformCommunity(PolymorphicModel): # previously called "Community"
- community: ForeignKey(Community)

SlackCommunity(PlatformCommunity)
- team_id: CharField(unique)

DiscordCommunity(PlatformCommunity)
- team_id: CharField(unique)

PlatformAction(BaseAction)
- community = models.ForeignKey(CommunityPlatform, models.CASCADE)
```

This should be merged with Metagov PR: https://github.com/metagov/metagov-prototype/pull/77

### Slack Installation Sequence Diagram

Metagov sequence diagrams - bot installation

---

- [x] "Add to Slack" on PolicyKit login page should hit metagov to perform installation of Metagov bot. (PolicyKit does not have its own Slack Client ID)
- [x] "Sign in with Slack" on PolicyKit login page should hit metagov to perform login. (PolicyKit does not have its own Slack Client ID)
- [x] PolicyKit should continue to be able to make requests on behalf of the installing user. (Installer must be admin). For now, just continue to let PolicyKit get those user auth tokens and store them in the SlackUser records. Later, add the necessary token overrides into the Metagov plugin, so that PolicyKit never has any auth tokens.
- [x] Replace "action" listener in PolicyKit integration with a receiver that gets webhook events from Metagov, and creates PlatformActions from them
- [x] Modify all `execute` and `revert` and `notify` functions to make requests to Metagov, not directly to Slack.
- [x] Create parent community model (see: https://github.com/amyxzhang/policykit/issues/384) and ~~write migrations for it~~ reset migrations.
- [x] Update Metagov Settings page to ~~hide Slack configurations~~ make Slack config read-only
- [x] Make it so Metagov creates the unique Community identifier itself (use a hash of the database key), so Drivers don't have to deal with this nonsense
- [x] Update Slack documentation in PolicyKit readthedocs
- [ ] Include auth flow diagram somewhere

TODOs
[...]

PR #436, opened 2021-07-13
Rename community notify_action -> initiate_vote, add post_message, fix Slack bugs

mashton:
Closes https://github.com/amyxzhang/policykit/issues/413

1. Rename `notify_action` to `initiate_vote`. The function signatures are unchanged.
2. Add method `post_message` to all communities. The common-ish API is `post_message(text, users, channel)` and each platform can have additional optional params on top of that.
3. Update all the sample policies.
4. Fix a few bugs in the Slack integrator that I found while doing this. The bug fixes are totally unrelated to the above items 🙈.  all bugs that I had introduced with https://github.com/amyxzhang/policykit/pull/408 .

### DIFF
--- a/docs/source/policy_model.rst
+++ b/docs/source/policy_model.rst
@@ -23,7 +23,7 @@ Check
 Notify
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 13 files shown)


mashton:
https://github.com/amyxzhang/policykit/pull/436/commits/59d1efbbba461b16f476380fe505823815da4d32 makes Slack and Discord backwards compatible for now, because why not. That way when we pull in this change elsewhere it wont immediately break all the voting policies.

amyxzhang:
looks good! thanks!

PR #437, opened 2021-07-14
Fix issue where celery runs old version of code

mashton:
I was running into an issue where celery was running pending tasks with an old version of the polciyengine code. Typically `systempctl restart celery` is required to get the updates, but for some reason that wasn't working. Even stopping and starting celery didn't work. This is the change that ultimately fixed it. I'm not sure why I'm only hitting this now. left a comment in there for future reference


amyxzhang:
hm weird, ok.

Issue #384, opened 2021-05-17
Support multi-platform communities


mashton:

mashton:
@amyxzhang @rickygv99 what do you think of this data model?


```
Community(Model):
- id: AutoField(unique)
- metagov_slug: CharField(unique)  # identifies this Community in Metagov

CommunityPlatform(PolymorphicModel):
- community: ForeignKey(Community)

SlackCommunityPlatform(CommunityPlatform)
- team_id: CharField(unique=True
- bot_id: CharField()

DiscordCommunityPlatform(CommunityPlatform)
- team_id: CharField(unique=True)

DiscourseCommunityPlatform(CommunityPlatform)
- team_id: CharField(unique=True)
```

The team_id is unique per platform, that's why it's defined in each subclass (this is how it's set up currently).

amyxzhang:
Overall good. I think the naming is a little wordy, so I would move it to be PlatformCommunity and then SlackCommunity, DiscordCommunity, etc.

mashton:
The renaming of `Community->PlatformCommunity` (and then a NEW model `Community`) was a lot harder thatn I expected @amyxzhang because Django migration system won't easily consider that to be a "model rename." I'd need to create a custom migration script to get it right and port everything over.
So, I gave up after a while and proceeded with `Community` and `ParentCommunity` instead. Is that OK? The model name is not user-facing.

amyxzhang:
That's fine by me. Are we that worried about db migrations at this point? If it's time-consuming, I'd rather just save the time and remake the DB.

mashton:
At this point there are a few instances of PolicyKit+Metagov, so each of us will have to do some form of manual data migration anyway so that we don't lose the PK Policies and the MG Plugin Community configuration (like stored API keys and so on). But yeah with this big of a change, I think you're right we should probably just reset "Django migrations" and perform the _actual_ data migration via custom scripts of downloading and uploading the data. Or else we will have a very complicated migration history in the code base.

amyxzhang:
I'm also happy to wait until we've accumulated all the major DB changes we plan to make in the near term and do a big reset all in one go.

mashton:
@amyxzhang here's an edge case I hit recently with https://github.com/amyxzhang/policykit/pull/438:

* User X goes to PolicyKit.org and installs PolicyKit to Slack workspace A. PolicyKit Community 001 is created.
[...]

PR #438, opened 2021-07-19
Add Metagov integration management UI

mashton:
This PR improves the UI for managing Metagov plugins. I'm calling them "integrations" in PolicyKit.

Things to note:
* Users that don't have perm `metagov.can_edit_metagov_config` can see everything (EXCEPT the current config key-values, since they may contain API keys). However, when they try to enable/disable plugins it will be rejected. There's a lot more we could do with this in the future (like allow users to "propose" an action that enables a plugin).
* **Slack** is not included in the dropdown for "Add integrations..." because we need full support for multi-platform comms before we can make this user experience clean (See https://github.com/amyxzhang/policykit/issues/384). Technically we could redirect users to auth Slack here and link the community, but the user would still need to log in and out to access "PolicyKit-for-discord" and "PolicyKit-for-slack." There are also some unhandled issues: (1) policykit hard-codes the oauth flow to redirect to /login, just need to update it to go to /main/settings when installingn to an existing comm, (2) metagov slack auth endpoint performs re-installation to existing community if one exists that already has slack plugin enabled with a matching team_id, which is NOT what we want when installing slack to an _existing_ community. figure out a better workflow.

----

Screen Shot 2021-07-19 at 10 30 36

Screen Shot 2021-07-19 at 10 30 23

Screen Shot 2021-07-19 at 10 01 00


amyxzhang:
Really exciting!! This looks good.

amyxzhang:
one thing I forgot to mention - I'd like to eventually move this out of under the user account drop down, to on the community's home page, say above the feed, or as a dropdown/gear image next to the community's name.

PR #444, opened 2021-07-21
Multi-platform communities, part 1

mashton:
Beginning support for multi-platform communities. I have this working for Discord + Slack. Because Discord isn't a Metagov plugin, yet, it has to be added in a specific order (Create PK Community using Discord, THEN add the Slack integration). After that, the user still has to log in and out to access the distinct DiscordCommunity and SlackCommunity, which still have separate policies and everything. The only thing they DO share is the integrations. So if I add an OpenCollective or Loomio integration in one community, it will also be present in the other one.

This also implements some better error handling logic, as specified in # 2 here https://github.com/amyxzhang/policykit/issues/384#issuecomment-882638075.

MUST be deployed with https://github.com/metagov/metagov-prototype/pull/101

## Walkthrough Discord+Slack example with screenshots
1. Install PolicyKit to Discord server "miri's server"
2. Sign in with Discord to "miri's server"
3. Go to the Settings page and enable SourceCred
4. Go to the Settings page and enable Slack
  Screen Shot 2021-07-21 at 15 55 51
  ...give my consent to slack workspace "test workspace"...
  Screen Shot 2021-07-21 at 16 12 29
  ...select a starterkit for my new SlackCommunity ...
   Screen Shot 2021-07-21 at 16 12 40
  ...redirected back to settings page for "miri's server"...
    Screen Shot 2021-07-21 at 16 23 24
5. Sign out

6. Sign in with Slack org "test workspace"

7. Go to Settings page. All the Plugins that are enabled for the DiscordCommunity are also enabled for this SlackCommunity!
   Screen Shot 2021-07-21 at 16 31 12

8. In the database, the SlackCommunity and DiscordCommunity have the same parent Community.

### DIFF
--- a/policykit/policyengine/views.py
+++ b/policykit/policyengine/views.py
@@ -30,9 +30,7 @@ def authorize_platform(request):
     platform = request.GET.get('platform')
     if not platform or platform != "slack":
         return HttpResponseBadRequest()
-    from policyengine.models import Community
-    new_community = Community.objects.create()
-    url = construct_authorize_install_url(request, integration=platform, community=new_community)
+    url = construct_authorize_install_url(request, integration=platform)
     return HttpResponseRedirect(url)
 
 @login_required(login_url='/login')
@@ -168,10 +166,10 @@ def settings_page(request):
 
         # FIXME(#384) support multi-platform communities. Here we're hiding slack from the "Add Integrations" dropdown,
         # because we don't have support for multiple CommunityPlatforms connected to one Community.
-        disabled_integrations_mod = [(k, v) for (k,v) in disabled_integrations if k not in ["slack"]]
+        # disabled_integrations_mod = [(k, v) for (k,v) in disabled_integrations if k not in ["slack"]]
 
         context["enabled_integrations"] = enabled_integrations.items()
-        context["disabled_integrations"] = disabled_integrations_mod
+        context["disabled_integrations"] = disabled_integrations
 
     return render(request, 'policyadmin/dashboard/settings.html', context)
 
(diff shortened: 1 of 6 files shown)


policykit/policyengine/views.py
```diff
@@ -242,6 +240,11 @@ def disable_integration(request):
     if community.platform == name:
         return redirect("/main/settings?error=not_permitted")
 
+    # Temporary: disallow disabling the plugins that have a corresponding PlatformCommunity.
+    # We would need to delete the SlackCommunity as well, which we should show a warning for!
+    if community.platform == "slack":
+        return redirect("/main/settings?error=not_permitted")
```
mashton, line 246: Ok, this is another edge case I'm putting off for now. 😓 

The edge case is:
1. User installs PolicyKit to Discord from /login page, which creates a new Community and a new DiscordCommunity
2. User adds Slack integration from /settings page, which creates a new SlackCommunity linked to an existing Community
3. User clicks "Delete (Slack) Integration" from /settings page while logged in with Discord.

I didn't want to just do `SlackCommunity.delete()` because the user could unintentionally lose tons of data. For now I just disallowed deleting the integration. fyi @amyxzhang

amyxzhang:
This looks good! In the future (this is more regarding the UI which should be tinkered with after we figure out the backend changes I think), I think I would want the "governed" platforms (the integrations with a corresponding Community object) to maybe have a different placement/experience than the other integrations. But I don't think we should prioritize that right now.

policykit/policyengine/views.py
amyxzhang: Yeah makes sense in the future to put a major warning here.

mashton:
yes definitely agreed @amyxzhang !

PR #446, opened 2021-07-23
Use a single Policy model for Platform policies and Constitution policies

mashton:
* Combine PlatformPolicy and ConstitutionPolicy into a single model called Policy
  * custom migration file ensures that policies arent lost, I tested on my server. it also downloads all policies to text files as backup.
* Require policies to have names, and show an error to the user if save is unsuccessful (So they don't lose their work!)
* Some fixes to starterkit constitution policies that were calling `action.execute()` on pass, which would lead to double execution since we do it from the engine


policykit/policyengine/models.py
```diff
@@ -1176,81 +1200,46 @@ class BasePolicy(models.Model):
...
     is_active = models.BooleanField(default=True)
     """True if the policy is active. Default is True."""
 
     modified_at = models.DateTimeField(auto_now=True)
     """Datetime object representing the last time the policy was modified."""
 
+    # TODO(https://github.com/amyxzhang/policykit/issues/341) add back support for policy bundles
+    bundled_policies = models.ManyToManyField("self", blank=True, symmetrical=False, related_name="member_of_bundle")
+    """Policies bundled inside this policy."""
+
```
mashton, line 1218: Policies can have other policies within them, and they can also be part of a policy bundle. That's why this is an asymetrical ManyToMany field. It can be used like this:

```
p = Policy.objects.all()[0]
p1 = Policy.objects.all()[1]

p.bundled_policies.add(p1)
p.bundled_policies.count() # 1, it has one bundled policy
p.member_of_bundle.count() # 0, it is not a member of any bundles

p1.bundled_policies.count() # 0, it doesn't have any bundled policies
p1.member_of_bundle.count() # 1, it is a member of one bundle

```

policykit/policyengine/models.py
```diff
@@ -1148,9 +1162,26 @@ def save(self, *args, **kwargs):
 
         super(PlatformActionBundle, self).save(*args, **kwargs)
 
+class PlatformPolicyManager(models.Manager):
+    def get_queryset(self):
+        return super().get_queryset().filter(kind=Policy.PLATFORM)
+
+class ConstitutionPolicyManager(models.Manager):
+    def get_queryset(self):
+        return super().get_queryset().filter(kind=Policy.CONSTITUTION)
```
mashton, line 1171: this lets you do `Policy.platform_policies.all()` and `Policy.constitution_policies.all()`

policykit/policyengine/tests/test_models.py
```diff
@@ -134,7 +134,8 @@ def setUp(self):
         self.data_store = DataStore.objects.create(
```
mashton, line 134: This test file doesn't work for me locally it fails with errors about Permissions missing. I guess it doesn't get run with `python manage.py test` for some reason. Do you use this locally @rickygv99 or is it just old?

policykit/integrations/slack/models.py
```diff
@@ -423,37 +422,21 @@ class Meta:
...
-                p.success = policy.success
-                p.fail = policy.fail
-                p.description = policy.description
-                p.name = policy.name
-
-                proposal = Proposal.objects.create(author=None, status=Proposal.PASSED)
-                p.proposal = proposal
-                p.save()
+            p = Policy()
+            p.kind = Policy.CONSTITUTION if policy.is_constitution else Policy.PLATFORM
```
mashton, line 426: @rickygv99 heads up this is likely going to conflict with your starterkit work. I think you were planning to remove the `GenericPolicy` construct, and just have start kits create new `ConstitutionPolicy` and `PlatformPolicy` records. If _this_ PR goes in, that would change yet again so that starterkits simply create `Policy` records. The field `kind` designates whether the policy is a constitution or platform policy.

amyxzhang:
Looks great, thank you!

Issue #341, opened 2021-04-15
Add documentation for "Add to Policy Bundle"


mashton:
There's no documentation in https://policykit.readthedocs.io/en/latest about policy bundles. We should add some documentation, and maybe an explanation in the authoring interface about what a policy bundle is and how it behaves.

mashton:
The point of policy bundles is to chain rounds of voting, because each Policy can only perform one vote. As soon as the policy `check` returns PASSED or FAILED, the policy execution ends.

PR #448, opened 2021-08-04
Patch StarterKit script

mashton:
Patch the StarterKit which was broken by https://github.com/amyxzhang/policykit/pull/446. That PR removed the `ConstitutionPolicyBundle` and `PlatformPolicyBundle` models, which were both replaced by the `Policy` model which can be a bundle / be bundled. As a result, the "can add" and "can view" `Permission`s for those two models no longer exist. I'm removing them here just to get that starterkit script working again.

We don't support Policy bundles in the UI yet anyway (https://github.com/amyxzhang/policykit/issues/341). But when we do add back support for it, we'll need to consider whether that action ("add to bundle") should have a separate permission attached to it.


amyxzhang:
Merging now!

Hmm I don't think we need a separate permission for bundles as adding to a bundle doesn't seem different to me than just creating a new action/policy.

PR #452, opened 2021-08-11
Enable GitHub integration through Metagov

mashton:
Enable GitHub integration through Metagov. The installation view is entirely copied from the Slack installer view. Look into consolidating this logic into a "core" PolicyKit view.

I tested in a policy like this:
```
response = metagov.perform_action("github.method", {"method": "GET","route":"/repos/{owner}/{repo}/issues", "repo":"myrepo"})
debug(str(response))
```

In order to change the repos you have access to, you currently need to delete and re-enable the integration in PolicyKit, which is not ideal.

### DIFF
--- a/policykit/integrations/github/__init__.py
+++ b/policykit/integrations/github/__init__.py
@@ -1,0 +1,1 @@
+default_app_config = 'integrations.github.apps.GithubIntegrationConfig'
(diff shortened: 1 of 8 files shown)


amyxzhang:
Looks good to me, I'm going to merge it.

Issue #440, opened 2021-07-20
Propose new API for performing x-platform actions/processes


mashton:
The current authoring API is not conducive to performing actions and votes on multiple platforms. This issue is to propose an alternative API that will sit on top of the new multi-platform data model https://github.com/amyxzhang/policykit/issues/384 

### Requirements
* As a policy author, I should be able to perform actions on any platform that is hooked up to my community
  * There shouldn't need to be a distinction between performing "actions that are governable" vs just "actions." For example `SourceCred.get_user_cred(...)` and `Slack.rename_channel(...)` could have the same syntax even though one is a governable PlatformAction and the other is just an api call to metagov
* As a policy author, I should be able to perform *one* vote on any voting-enabled-platform that is hooked up to my community
  * I should also be able to get the results of the vote at any time, with the users attached to the votes if available.
* As a policy author, I should be able to *propose* a new PlatformAction (aka "put it in the queue")
* As a policy author, I should NOT be able to access any actions/policies/data associated with other communities
* The syntax should be conducive to adding typehints/code completion.

### First pass
first attempt, rewriting one of our existing policies. This one is a "trigger" policy not a governing-action policy. 
```py
import datetime
import Discourse  # Class that gets all the available actions and processes dynamically added to it?
import OpenCollective
import TriggerEvent # Analog to PlatformAction (irreversible event)
import EvaluationState # TBD???

def filter(event: TriggerEvent):
    return event.platform == "opencollective" and event.type == "expense_created"


def notify(event: TriggerEvent):
    response = Discourse.initiate_vote(title="vote yes or no on expense", category=8, details="etc etc")
    poll_url = response.poll_url  # there are attributes on the response (?)

    expense_id = event.id  # there are attributes on the event, rather than just a json blob (?)
    message = f"Thank you for submitting an expense. Please vote at {poll_url}."
    response = OpenCollective.create_comment(expense_id=expense_id, raw=message)


def check(event: TriggerEvent, state: EvaluationState):
    yes_votes = state.get_yes_votes() # are these models? objects? classes? how do we access voter list?
    no_votes = state.get_no_votes()
    if state.get_time_elapsed() < datetime.timedelta(minutes=5):
	    return None # indicates pending..
    
    approved = yes_votes > no_votes
    # approve/reject the expense
    action = "APPROVED" if approved else "REJECT"
    OpenCollective.process_expense(expense_id=event.id, action=action)
    # notify the users
[...]

PR #449, opened 2021-08-06
Link Proposal to Policy, refactor engine, reset migrations

mashton:
##### TO DO:
- [x] Update starter kit policies and update documentation. **Wait until Ricky's starter kit change goes in.**
- [x] Add test for celery task func
- [x] Update documentation around PolicyEvaluation
- [x] Update sample policies. There was an api change for `get_yes_votes` etc, since "proposal" is no longer a thing.
- [x] Move `community_post` to PolicyEvaluation instead of PlatformAction

---

### Data model changes
* Made `BaseAction` concrete. Implementing a migration for this change [is complex](https://tech.agilitynerd.com/django-migrate-abstract-concrete-base-class.html) so we decided to reset migrations for this PR instead.
* Changed `Proposal` model. Proposal now represents an evaluation of a single action against a single policy. It has foreign key references to `BaseAction` and `Policy`.
* Remove `MetagovProcess` from metagov integration. This model is no longer needed, since we can store the governance process URL and JSON data on the `Proposal` record. The authoring API for initiatating generic metagov processes has not changed.
* Some changes to django_db_logger so logs are tied to specific evaluations
* Removed `action_codename` in favor of `action_type` and make it a calculated property so each model doesn't need to add it manually

### Refactor changes
* Move Policy Engine logic out of `views.py` into `engine.py`
* Consolidate the policy execution code. Instead of having separate functions for `check_policy` and `initialize_policy` etc, we have one `exec_code_block` function that's used for all blocks. All blocks get the same context available to them.

### Changes to policy authoring API
* Implements proposal from https://github.com/amyxzhang/policykit/issues/440#issuecomment-891940276
* Authors now have access to all available PlatformCommunity instances as in-scope variables `slack`/`discord`/`discourse` etc. They are still accessible via `action.community` and `policy.community` too, so this isn't a breaking change.
* `debug("..")` is removed in favor of `logger` which can be used like a typical Python logger. This allows authors to log at different levels, and we can eventually add filtering on the log page for that. **This is a breaking change** for policies that use "debug".

### UI changes
Now you can see policy data in the "pending actions" and "logs" views:
updated_pendingevaluations
updated_logtable

### Random thoughts
* Unhandled edge case: what if a user receives `can_execute` permissions while an action is in pending state? we don't re-check that with each evaluation
[...]

PR #454, opened 2021-08-13
Cross-platform policy authoring

mashton:
Implements proposal from https://github.com/amyxzhang/policykit/issues/440#issuecomment-891940276

### Changes to policy authoring API
* Authors now have access to all available PlatformCommunity instances as in-scope variables `slack`/`discord`/`discourse` etc. They are still accessible via `action.community` and `policy.community` too, so this isn't a breaking change.
* `debug("..")` is removed in favor of `logger` which can be used like a typical Python logger. This allows authors to log at different levels, and we can eventually add filtering on the log page for that. **This is a breaking change** for policies that use "debug".


policykit/policyengine/engine.py
```diff
@@ -0,0 +1,309 @@
...
+        action (BaseAction): The action that triggered this policy evaluation.
+        policy (Policy): The policy being evaluated.
+        slack (SlackCommunity)
+        discord (DiscordCommunity)
+        discourse (DiscourseCommunity)
+        reddit (RedditCommunity)
+        metagov (Metagov): Metagov library for performing enabled actions and processes.
+        logger (logging.Logger): Logger that will log messages to the PolicyKit web interface.
+
+    """
```
mashton, line 34: This class isn't actually exposed to the policy author, but we use it to document all the in-scope variables. Any attributes on this class will be available to the policy author.

policykit/policyengine/engine.py
```diff
@@ -0,0 +1,309 @@
...
+            db_logger, {"community": proposal.action.community.community, "proposal": proposal}
+        )
+
+        from policyengine.models import Community, CommunityPlatform
+
+        parent_community: Community = self.action.community.community
+
+        # Make all CommunityPlatforms available in the evaluation context
+        for comm in CommunityPlatform.objects.filter(community=parent_community):
+            setattr(self, comm.platform, comm)
```
mashton, line 50: This is the key change that makes policies now be "cross platform." For example in a Slack policy I could say `discord.post_message(...)` or `discord.initiate_vote(...)`, if I had a discord community connected to my slack community.

policykit/policyengine/engine.py
```diff
@@ -0,0 +1,309 @@
...
+
+    Attributes:
+        proposal (Proposal): The proposal representing this evaluation.
+        action (BaseAction): The action that triggered this policy evaluation.
+        policy (Policy): The policy being evaluated.
+        slack (SlackCommunity)
+        discord (DiscordCommunity)
+        discourse (DiscourseCommunity)
+        reddit (RedditCommunity)
+        metagov (Metagov): Metagov library for performing enabled actions and processes.
```
mashton, line 31: This is staying the same, for now. I'll do a follow up PR to change this, so we have light-weight integrations (for Loomio, Open Collective, SourceCred) instead of a generic "metagov"  class. I'm still not 100% sure what that API will look like for starting voting processes and performing actions, but that's the next think I'll tackle. @amyxzhang

mashton:
closing and moving commit https://github.com/amyxzhang/policykit/pull/454/commits/60dbf5c2f0c13216ca12418ed5ca67bd764a6b19 to  https://github.com/amyxzhang/policykit/pull/449

PR #469, opened 2021-09-20
Fix Slack voting from trigger policies

mashton:
Fix bug where votes initiated from trigger policies were not counting any of the votes cast (fixes seminar-planning policy)


amyxzhang:
:) Good to see you back @mashton

PR #473, opened 2021-09-28
Fix bug where users could edit policies and roles of other communities

mashton:
Fix a bug where users could manipulate IDs in the query string (like `https://policykit.org/main/editor/?type=Platform&operation=Change&policy=3`) to view and edit policies from other communities. This updates all views to check that the document being requested belongs to the community that the user is logged in to.


amyxzhang:
Thanks for catching this!

PR #477, opened 2021-10-12
Add Open Collective integration with triggers

mashton:
* Add OpencollectiveCommunity, OpencollectiveUser, and TriggerActions for expense events
* Small refactor: make it so that integrators write their own views for `enable_integration` and `disable_integration` since that will involve created/deleting CommunityPlatform instances. For the remaining "generic" plugins (Loomio, Mailgun, etc.) the `metagov` app handles the enable/disable view.
* Add 500/404/403 pages
* Deployed on https://policykit.metagov.org/login feel free to log in with Test workspace and play around @amyxzhang


amyxzhang:
Thanks, looks good!

PR #533, opened 2021-12-27
Fix bugs in safe restricted-python evaluation

mashton:
* Use safe getattr method to restrict access to `_`-prefixed attributes
* Add guards for iterators to make lists and loops work
* https://github.com/amyxzhang/policykit/pull/533/commits/8516a3def1f94de7f4455347ccce2aa7edd7e097 Fix bug where `strftime` was raising an exception https://github.com/amyxzhang/policykit/issues/534
* https://github.com/amyxzhang/policykit/pull/533/commits/8516a3def1f94de7f4455347ccce2aa7edd7e097 Include line numbers for syntax errors in web log viewer:
    Screen Shot 2021-12-27 at 22 29 55
    Screen Shot 2021-12-27 at 22 29 22

### DIFF
--- a/policykit/policyengine/safe_exec_code.py
+++ b/policykit/policyengine/safe_exec_code.py
@@ -10,13 +10,12 @@ import base64
 import itertools
 import json
 
-
-BUILTINS = {
+policykit_builtins = {
     # see: https://restrictedpython.readthedocs.io/en/latest/usage/policy.html#predefined-builtins
     **safe_builtins,
-    # access to standard modules like math, random, string and set.
+    # standard modules like math, random, string and set.
     **utility_builtins,
-    # other permitted modules
+    # other modules that are permitted in policies, like datetime and json
     "datetime": datetime,
     "base64": base64,
     "itertools": itertools,
@@ -32,7 +31,7 @@ STATIC_GLOBAL_VARIABLES = {
 
 class OwnRestrictingNodeTransformer(RestrictingNodeTransformer):
     def visit_Import(self, node):
-        self.error(node, "Import statements are not allowed.")
+        raise SyntaxError("Import statements are not allowed.", ("<user_code>", node.lineno, node.col_offset, ""))
 
     visit_ImportFrom = visit_Import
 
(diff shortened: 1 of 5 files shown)


policykit/policyengine/safe_exec_code.py
```diff
@@ -36,6 +37,17 @@ def visit_Import(self, node):
     visit_ImportFrom = visit_Import
 
 
+def _hook_writable(obj):
+    """Only allow writing to lists and dicts."""
+    # if obj.__class__.__name__ == "MY_CLASS_NAME":
+    #     return obj
```
mashton, line 43: If we ever want to support assigning attributes to certain objects (like, `proposal.status = FAILED` we could permit those objects here

policykit/policyengine/tests/test_exec_code.py
```diff
@@ -40,6 +57,28 @@ def test(x, name="alice"):
             execute_user_code(example, "test", 5)
         self.assertTrue("Import statements are not allowed" in str(cm.exception))
 
+    def test_execute_unsafe_access(self):
+        """Test that getattr is restricted"""
+        example = """
+def test(inst):
+    return inst._private_attr
```
mashton, line 64: The `safer_getattr` guard restricts users from accessing attributes that are prefixed with `_`. This is helpful because we can use this prefix to restrict any functions on communities that we don't want to expose to policy authors (I've already been doing this by convention, but now it's actually enforced). @amyxzhang

policykit/policyengine/safe_exec_code.py
amyxzhang: I think definitely proposal.status should be allowed to be assigned. Any others that come to mind?

policykit/policyengine/tests/test_exec_code.py
amyxzhang: Awesome!!

amyxzhang: And these attributes won't show up in the autocomplete as well, right?

amyxzhang:
Looks great! Feel free to merge!

policykit/policyengine/safe_exec_code.py
mashton: No, right now I can't think of any reason to allow assigning attributes on objects. So it is completely restricted for now.

policykit/policyengine/tests/test_exec_code.py
mashton: Yes they don't show up in autocomplete, because we filter them out [here in the autocomplete module](https://github.com/amyxzhang/policykit/blob/6be3d0da3100c104e6d72d646b6b9ecd4c405760/policykit/policyengine/autocomplete.py#L58)

Issue #445, opened 2021-07-22
Search for users in user list


edsaperia:
When I am trying to select a user so that I can grant a role to that user, the dropdown list of users is very long and not in an obvious order, so it's hard to find the user I want! It would be good to be able to search the list.

This functionality may be useful for any situation where I am choosing something from what may be a long list, so maybe see if you can add it to other dropdowns too!

Issue #515, opened 2021-12-08
Update dashboard user list to include platform (for multi-platform communities)


mashton:
The PolicyKit dashboard hasn't been updated since we added support for multi-platform communities. Some things should be updated, such as:
* In the User list, show labels indicating which Platform the user belongs to. For example there might be users `alice123 [discourse]` and `Alice [slack]` etc. This should be updated in the user dropdown lists when assigning roles, too. (Related improvement to user list: https://github.com/amyxzhang/policykit/issues/445)
* On the dashboard, show which platforms the community is connected to.

PR #536, opened 2021-12-28
Use bootstrap-select for searchable user list organized by platform

mashton:
Closes https://github.com/amyxzhang/policykit/issues/445
This is the same style select we use for the trigger selection on the policy editor.

Screen Shot 2021-12-28 at 09 23 06


mashton:
Related issue https://github.com/amyxzhang/policykit/issues/515 is for separating users by platform on the dashboard user list

edsaperia:
Beautiful, thank you!

PR #559, opened 2022-02-25
Fix bug where only 1000 discord members are inserted into db

hozzjss:
- this patch fixes a bug where only servers with 1000 members or less would have access to pk, the rest of the 1000 aren't even recognized by pk


amyxzhang:
Looks great, thanks!

PR #560, opened 2022-03-06
Added docker integration

PatelFarhaan:


shaunagm:
Hi @PatelFarhaan, can you add some documentation/instructions for how to use this integration? I'm not familiar enough with docker to be able to test this without docs, and we should probably have the docs regardless.

amyxzhang:
@PatelFarhaan I believe before I can merge this, you still need to add docs? Second, I see a unit check that's failed - can you look into that?

PatelFarhaan:
@shaunagm @amyxzhang I have added the docs.

saulshanabrook:
Note that some docker support has landed in https://github.com/policykit/policykit/pull/631. It is only for development at the moment, not production use cases.

PR #575, opened 2022-12-05
feat: copy policy

orzubalsky:
Add a `copy` method to the`Policy` model, to be used to generate copies from template policies.

- `policy_instance.copy()` makes a deep copy of the policy, maintaining m2m relationships, and making new copies of policy variables.
- `policy_instance.copy()` must be called with a community argument, to which the copy policy will be related.
- Copies of policy variables are created with value set to the default value.
- Updates are supported by additional unit tests

### DIFF
--- a/policykit/policyengine/models.py
+++ b/policykit/policyengine/models.py
@@ -861,6 +861,40 @@ class Policy(models.Model):
         """True if the policy is part of a bundle"""
         return self.member_of_bundle.count() > 0
 
+    def copy(self, community = None):
+        """Make a copy of the policy object and assign to a new community"""
+        if not community:
+            raise Exception("Community object must be passed")
+
+        from copy import deepcopy
+
+        # Make a copy of the whole objet
+        new_policy = deepcopy(self)
+
+        # Generate a new id
+        new_policy.pk = None
+
+        # Assign copy to another community
+        new_policy.community = community
+
+        new_policy.save()
+
+        # Copy ActionType relationships
+        new_policy.action_types.set(self.action_types.all())
+
+        # Remove existing Policy Variable relationships
+        new_policy.variables.set([])
+
+        # Make copies of related PolicyVariables
+        for variable in self.variables.all():
+            new_variable = deepcopy(variable)
+            new_variable.pk = None
+            new_variable.policy = new_policy
+            new_variable.value = variable.default_value
+            new_variable.save()
+
+        return new_policy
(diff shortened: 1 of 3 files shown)


amyxzhang:
Question - to clarify, right now "template policies" are just policies, right? As in these are just like any other policy tied to a community and currently being used by that community? If so, we would want to set the new policy variables' values to be the default value instead of copying what was the original policy's value? Otherwise, we may leak private info about the original community.

Alternatively, we could decide now to have a concept of a "template policy" - it is like a policy in every way except it does not have a community and any policy variables don't have a value set. It could also have metadata that speaks to how it was created/shared, imagining a future UI where users can publish policies as template policies for others to use and customize.

amyxzhang:
I agree it's lower priority... though I worry we would need to re-do a bunch of things as a result. Conceptually, it doesn't really make sense to copy an existing community's policy to create a new policy. Also, practically, we would author our OC policy probably on MetaGov's PK community but then it prevents MetaGov from being able to adjust that policy for ourselves (outside of the variables), since we know other communities will be copying from it. The other option is to create a dummy PK community hooked up to a dummy OC and slack, which also seems not ideal.

Would it be hard to do to create the concept of a template policy in the backend for now? Can we simply use the existing Policy model somehow, maybe just a new flag/community=None and some added variables and functions...? This copy functionality then would only be for policies that are templates. And we'd want a function for create a new template policy from an existing community policy.

orzubalsky:
Adding an `is_template` flag to the `Policy` model and updating the `copy()` method accordingly sounds like a good solution. How about I add that in a separate pr?

amyxzhang:
That sounds good! In which case I think this PR is good to go.

amyxzhang:
That sounds good! In which case I think this PR is good to go.

policykit/policyengine/models.py
```diff
@@ -861,6 +861,40 @@ def is_bundled(self):
...
         return self.member_of_bundle.count() > 0
 
+    def copy(self, community = None):
+        """Make a copy of the policy object and assign to a new community"""
+        if not community:
+            raise Exception("Community object must be passed")
+
+        from copy import deepcopy
+
+        # Make a copy of the whole objet
```
amyxzhang, line 871: minor typo on object

amyxzhang (approved):

gigxz (approved):

PR #577, opened 2022-12-09
Feature/nocode mvp UI

orzubalsky:
wip for discussion.

Preview at https://test.policykit.org/embed/opencollective/?source=3


orzubalsky:
Thanks, @amyxzhang. 

I removed all references to "open collective" in the urls, views, and templates and kept all of this UI flow under "embed". 

This flow only works with Policy objects that are `is_template=True`. There is one that you can use, policy `19`, although the code blocks don't have the variables' names in them yet. The first step of the UI flow is at: `https://test.policykit.org/embed/?source=19`.

The piece that's missing is the first step, involving logging in and getting permissions from both OC and Slack, and then creating a new PK community with a starterkit.

PR #578, opened 2022-12-12
Feat: policy can be designated as a template

orzubalsky:
Introduce the concept of policy as a template.

- Adds an `is_template` boolean field to the `Policy` model.
- Removes `community` relationship constraint from `Policy` model, making it nullable. 
- Adds validation check on `policy.save()`, ensuring that non-template policies must have a community assigned.
- Copying a policy to a community only works when source policy is a template.
- Adds a method to generate a template copy from an existing policy.
- Updates are supported by additional unit tests

Note: this pull request addresses the need expressed in the discussion regarding #575


amyxzhang:
This all looks good to me. I can go ahead and merge it in and deploy it in production. @orzubalsky is it already deployed in the test server?

PR #588, opened 2023-03-01
Prepping embed user flow for OC users

nickmvincent:
Adds several features to the embed flow:
- populate templates for embed flow
- select Slack channel by name instead of ID
- minor UI changes (not final: will be updated again based on design suggestion)


amyxzhang:
Sorry, was I supposed to look at this PR for merging or is it still WIP?

nickmvincent:
Ah, sorry, wasn't clear!

Ready for merge, I think, but assuming the visuals will still change quite a bit.

policykit/policyengine/views.py
```diff
@@ -824,49 +824,177 @@ def document_action_recover(request):
...
+    Get policy by name.
+    """
+    from policyengine.models import Policy
+    name = request.GET.get(key_name)
+    try:
+        return Policy.objects.get(name=name)
+    except Policy.DoesNotExist:
+        return(None)
+
+def embed_populate_templates(request):
```
amyxzhang, line 846: Yeah, I don't want to be creating one-off policies here. Also the policy engine should have no platform-specific logic. This should be a python script runnable from the command line using manage.py and stored in a platform-specific place.

policykit/policyengine/views.py
```diff
@@ -824,49 +824,177 @@ def document_action_recover(request):
...
+
+        PolicyVariable.objects.create(
+            name="pong_message", label="What to say in response to ping", default_value="pong", is_required=True,
+            prompt="What to say in response to ping", type="string", policy=policy)
+
+    return embed_select_template(request)
+
+
+
+def get_channel_options(community_id):
```
amyxzhang, line 913: same, there shouldn't be platform-specific logic in the policy engine.

policykit/policyengine/views.py
```diff
@@ -824,49 +824,177 @@ def document_action_recover(request):
...
+
+    slack_community = SlackCommunity.objects.get(community_id=community_id)
+    for channel in slack_community.get_conversations()['channels']:
+        if channel['is_channel']: # get only the "channels" (as opposed to group, im, mpim, private)
+            channel_options.append(
+                {'name': channel['name'], 'channel_id': channel['id']}
+            )
+    return channel_options
+
+def embed_select_template(request):
```
amyxzhang, line 928: with Leijie's changes, this code may need to be changed. we are planning to move away from "is_template=True" policies in general.

amyxzhang:
Ok I think there will need to be a couple changes. This may also need to wait until after Leijie's changes to how we think about policy templates.

PR #589, opened 2023-03-02
Draft Data models of No-code policies 

leijie-wang:
Users should specify three components to author a policy: `CustomAction`, `Procedure`, and `Execution`. 
1. `CustomAction`: ActionType (the GovernableAction it builds upon), a filter code block, and a community name.
2. `Procedure`: an abstract class. There are two procedures available now: `VotingProcedure` and `PermissionProcedure`.  
-  VotingProcedure: Because the initiate_vote and post_message function interfaces differ across integrations, we added subclasses in each integration, such as SlackVotingProcedure.
- PermissionProcedure: for each CustomAction that has a community name, create a new Permission object
3. `Execution`: ForeignKey to the Governable Action


amyxzhang:
Can you clean up the PR and remove all the files where you've made changes to the file permissions? That'll make it easier to see what you did, thanks!

PR #590, opened 2023-03-08
Draft Data models of No-code policies (clean version)

leijie-wang:
Users should specify three components to author a policy: `CustomAction`, `Procedure`, and `Execution`. 
1. `CustomAction`: ActionType (the GovernableAction it builds upon), a filter code block, and a community name. We create a new permission object for each new custom action instance with a non-empty community name.
2. `Procedure`: initialize, check, and notify code blocks, name of the procedure, and variables_dict used in the procedure 
3. `Execution`: ForeignKey to the Governable Action

I also started adding codes about the new frontend in this pull request.

### DIFF
--- a/policykit/policyengine/views.py
+++ b/policykit/policyengine/views.py
@@ -918,7 +918,10 @@ def design_custom_action(request):
     filter_parameters = {}
     new_actions = {}
     
+    new_actions = {}
+    
     for app_name, action_list in actions.items():
+        new_action_list = []
         new_action_list = []
         for action_code, _ in action_list:
             parameter = Utils.get_filter_parameters(app_name, action_code)
@@ -931,8 +934,19 @@ def design_custom_action(request):
             new_actions[app_name] = new_action_list
                 
 
+            parameter = Utils.get_filter_parameters(app_name, action_code)
+            # only show actions that have filter parameters
+            if parameter:
+                filter_parameters[action_code] = parameter
+                new_action_list.append((action_code, _))
+        # only allow apps that have actions with filter parameters
+        if new_action_list:
+            new_actions[app_name] = new_action_list
+                
+
     return render(request, "no-code/custom_action.html", {
         "trigger": trigger,
+        "actions": new_actions,
         "actions": new_actions,
         "filter_parameters": json.dumps(filter_parameters),
     })
@@ -953,9 +967,13 @@ def create_custom_action(request):
     from policyengine.models import CustomAction, ActionType, Policy, PolicyActionKind
 
 
+    from policyengine.models import CustomAction, ActionType, Policy, PolicyActionKind
+
+
     data = json.loads(request.body)
     action_type = data.get("action_type", None)
     if action_type:
+        # create a new CustomAction instance
         # create a new CustomAction instance
         action_type = ActionType.objects.filter(codename=action_type).first()
         is_trigger = data.get("trigger", "false") == "true"
@@ -973,12 +991,8 @@ def create_custom_action(request):
         new_policy = Policy.objects.create(
                 kind=policy_kind,
                 filter=custom_action.filter, 
-<<<<<<< HEAD
                 community=community,
                 is_active=False # do not activate the policy until the user has finished the flow
-=======
-                community=community
->>>>>>> 4a72375789ba68837c55ff917473cf6ce5d688ea
             )
         new_policy.action_types.add(custom_action.action_type)
 
@@ -1041,21 +1055,14 @@ def create_procedure(request):
     data = json.loads(request.body)
     template_index = data.get("template_index", None)
     policy_id = data.get("policy_id", None)
-<<<<<<< HEAD
     variables_data = data.get("data", {})
-=======
->>>>>>> 4a72375789ba68837c55ff917473cf6ce5d688ea
     if template_index and policy_id:
         procedure_template = Procedure.objects.filter(pk=template_index).first()
         new_policy = Policy.objects.filter(pk=policy_id).first()
         new_policy.initialize = procedure_template.initialize_code
         new_policy.check = procedure_template.check_code
         new_policy.notify = procedure_template.notify_code
-<<<<<<< HEAD
         procedure_template.create_policy_variables(new_policy, variables_data)
-=======
-        procedure_template.create_policy_variables(new_policy)
->>>>>>> 4a72375789ba68837c55ff917473cf6ce5d688ea
         new_policy.save()
         return JsonResponse({"status": "success", "policy_id": new_policy.pk})
     else:
@@ -1146,7 +1153,6 @@ def create_execution(request):
     else:
         return JsonResponse({"status": "fail"})
 
-
 @login_required 
 def policy_overview(request):
     trigger = request.GET.get("trigger", "false")
@@ -1170,10 +1176,7 @@ def create_overview(request):
         policy = Policy.objects.filter(pk=int(policy_id)).first()
         policy.name = data.get("name", "")
         policy.description = data.get("description", "")
-<<<<<<< HEAD
         policy.is_active = True
-=======
->>>>>>> 4a72375789ba68837c55ff917473cf6ce5d688ea
         policy.save()
         return JsonResponse({"policy_id": policy.pk, "policy_type": (policy.kind).capitalize(), "status": "success"})
     else:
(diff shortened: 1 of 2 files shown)


policykit/policyengine/models.py
```diff
@@ -1026,6 +1027,143 @@ def __init__(self, *args, **kwargs):
...
+
+        return {
+            "initial": self.initial_code,
+            "notify": self.notify_code,
+            "check": self.check_code,
+        }
+
+class Execution(models.Model):
+    
+    action = models.ForeignKey(GovernableAction, on_delete=models.SET_NULL)
```
amyxzhang, line 1149: Do we actually want this to be a foreignkey to an action? Do we want to create an action object whenever we run an execution? What would the purpose of that be (in terms of storing data we want to use later)? I think we want actions at the beginning because they form proposals, and we need the stored data both to carry out the procedure and to be able to execute the action (if need be) when passed. But I don't see the need to create an action object at execution time.

policykit/policyengine/models.py
```diff
@@ -1026,6 +1027,143 @@ def __init__(self, *args, **kwargs):
...
+    name = models.TextField(blank=True, default='')
+    """the name of the procedure if exists"""
+
+    initial_code = models.TextField(blank=True, default='')
+
+    notify_code = models.TextField(blank=True, default='')
+
+    check_code = models.TextField(blank=True, default='')
+
+    variables_dict = models.TextField(blank=True, default='')
```
amyxzhang, line 1128: I'm confused why we're storing this in a dict instead of just directly having a one-to-many to the PolicyVariables table?

policykit/policyengine/module.py
```diff
@@ -0,0 +1,76 @@
...
+)
+
+class PolicyFactory():
+
+    def create_builtin_procedure_templates():
+        """
+            Create procedute templates for a given platform
+        """
+
+        variables_for_simple_vote_procedure = {
```
amyxzhang, line 17: Can we move this into a Slack specific script? It's a one-off script that just needs to run once at database initialization. I think we have something like this already for loading into the db our default starterkits?

policykit/policyengine/models.py
```diff
@@ -1026,6 +1027,143 @@ def __init__(self, *args, **kwargs):
         self.label_suffix = ''
 
 
+class CustomAction(models.Model):
```
amyxzhang, line 1030: One thought - right now we still have the idea of Policy Templates via the Policy model (which has an is_template field). I wonder if we want to get rid of that and instead have a new PolicyTemplate class that can bundle an Action/CustomAction + a Procedure + an Execution. This would be useful in the case of say, being able to select from a set of templates (eg "Community vote on OC expenses on Slack" like how Nick is doing). This would fill everything in so the user doesn't have to go through the 3-stage process unless they want to edit something.

policykit/policyengine/models.py
amyxzhang: At the end of the day, the user is selecting from a menu of action types, and perhaps specifying parameters for that action type. It's similar to CustomAction in fact? We don't need to create the action object, we should just run the code to execute.

policykit/policyengine/module.py
```diff
@@ -0,0 +1,76 @@
+from django.db import models
+import json
+from policyengine.models import (
+    CommunityPlatform,
+    CustomAction,
+    Procedure,
+    Execution,
+)
+
+class PolicyFactory():
```
amyxzhang, line 10: shouldn't the policy factory be creating policies?

policykit/policyengine/models.py
amyxzhang, line 1030: I wonder if a better name for this is CustomActionType. Since it's not really an action.

policykit/policyengine/models.py
leijie-wang: Then we can keep a reference to the `ActionType` here and put the execution codes of each specific GovernableAction in an execution code block? When I was creating a reference to the GovernableAction, my intention was to have an organized way to store the execution parameters of each Governable action so that we can directly know who is the user, and what message is posted (compared to embedding all these parameters in the rather execution codes of different shapes). But I do not have a clear idea about what is the future use of these organized data and whether it is worth the cost.

policykit/policyengine/module.py
leijie-wang: Yes, indeed. I will pick a better name here

policykit/policyengine/models.py
leijie-wang: I like the idea! Right now, if we have already decomposed a policy template into three modules, the policy templates by the field`is_template` in the `Policy` seems repetitive; and we still would like to provide users with some complete policies

policykit/policyengine/models.py
amyxzhang: Hm yes, one way to do this is to create action objects in memory but never actually commit them to the database.

policykit/policyengine/models.py
amyxzhang: Yes, and a PolicyTemplate could include multiple CustomActions/Actions and multiple Executions as well (I think in the UI too, we want the ability to allow users to select multiple).

policykit/policyengine/models.py
amyxzhang: Alright

policykit/policyengine/models.py
amyxzhang: Shouldn't we distinguish Pass vs Fail as well in here?

policykit/policyengine/models.py
amyxzhang: Actually - if we are to have a PolicyTemplate class, why not store policy variables in there instead of in here? What if you want to use a policy variable in your filter step or execute step?

amyxzhang: Actually, that's not a good idea since we want Procedures to be able to exist and have PolicyVariables independent of any PolicyTemplate. Hmmmm so I guess we store PolicyVariables in Procedure for now and leave it at that.

policykit/policyengine/module.py
amyxzhang: well we do need a place to create Policy objects from these template classes, no?

policykit/policyengine/models.py
leijie-wang: I am thinking users can create one Execution instance for Pass and Fail respectively. Putting the PASS and FAIL code blocks together seems to assume that these two are quite related, and also introduces additional logic in the Execution implementation.

policykit/policyengine/models.py
leijie-wang: Yeah, I agree, since we still would like to keep the Procedure class an independent class and the sharing and creation of Procedures also go with their corresponding PolicyVariables

policykit/policyengine/module.py
leijie-wang: Yes, it is obviously a Slack procedure. My concern is actually about the bigger picture: I am not sure whether we would like to put in these integration-specific Procedure class, such as SlackProcedure. 

The first approach: Each defines one part of the procedure templates. The generic `Procedures` is responsible for defining the core procedure logic (such as how to determine the victory of a vote) and `SlackProcedure` for dealing with notifying users and initiating votes. 

The second approach: each defines the whole procedure but differs in how generalizable these procedures are. The generic `Procedures` define procedure templates that are applicable to all integrations, while `SlackProcedure` defines procedure templates that are only applicable to Slack.

policykit/policyengine/module.py
leijie-wang: I plan to use PolicyFactory for my creating these templates and putting the logic of creating policies from these templates in the `views.py` because it involves the interaction with the front end.

policykit/policyengine/module.py
leijie-wang: I think we would like to give users more freedom in creating policies and the first approach seems to assume the notification and initialization part of a procedure are less important, which restricts users' freedom in creating policies. 

But for the second approach, I also feel it is hard to define a prior whether a procedure is only applicable to this platform, as this may also depend on the similarity between this platform and that platform.

policykit/policyengine/models.py
amyxzhang: ok, then do we need a field specifying which one it is?

PR #595, opened 2023-05-16
Adding full support for early beta version of CollectiveVoice, using new no-code datamodels

nickmvincent:
This PR contains all the code needed (new views, new HTML templates, new backend code) to use the "CollectiveVoice" flow to create a Policy that allows for OpenCollective expense voting in Slack.

Conceptually, the current version of CollectiveVoice is a "constrained" version of no-code PolicyKit: the possible variations in event filtering and voting rules are constrained, in order to reduce the number of inputs needed to create a policy.


nickmvincent:
Closing, this is redundant with #598

PR #599, opened 2023-05-25
Given "CollectiveVoice" code its own app

nickmvincent:
In order to keep any code related to CollectiveVoice (views, urls, Slack or OpenCollective specific handling that *isn't* in an integration), etc. out of the core policyengine/ and policykit/ code, we move everything to a separate Django app called "policybuilding_apps". In theory, we could use this, or even use Django sub-apps in the future to support other specific use-cases like PolicyKit (e.g., a specific Discord <-> OC integration).

We create the app with `startapp` and add it to settings.py. We move all views and url code. So far, there are no models, so no migrations needed.


nickmvincent:
Reopening this from a separate feature branch again to avoid duplicate commits in history. Sorry!

PR #600, opened 2023-05-25
Give "CollectiveVoice" code its own Django app

nickmvincent:
In order to keep any code related to CollectiveVoice (views, urls, Slack or OpenCollective specific handling that isn't in an integration), etc. out of the core policyengine/ and policykit/ code, we move everything to a separate Django app called "policybuilding_apps". In theory, we could use this, or even use Django sub-apps in the future to support other specific use-cases like PolicyKit (e.g., a specific Discord <-> OC integration).

We create the app with startapp and add it to settings.py. We move all views and url code. So far, there are no models, so no migrations needed.


amyxzhang (approved):

amyxzhang:
Looks good! We may want to update the README and readthedocs at some point.

PR #624, opened 2024-04-29
implement the /api/members endpoint

ddouglascarr:
This PR implements the `/api/members` endpoint, as described in the api doc and in Slack discussions.

I started by copying the logic from `policyengine.views.role_action_users`. It remains mostly a straight copy from there, with a bit of tightening to prevent assigning roles across communities. I'm not 100% what the rules are there, so please check I haven't made it impossible to assign roles in related communities where it should work. I could not identify tests which covered that view, so I've written my own for the new one. 

This is the first DRF views for the project. Please don't hesitate to request changes if there's a different style or idiom that the project would like to use.

### DIFF
--- a/policykit/policyengine/api_views.py
+++ b/policykit/policyengine/api_views.py
@@ -1,0 +1,59 @@
+from rest_framework.decorators import api_view, permission_classes
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+from rest_framework.exceptions import NotFound 
+from django.contrib.auth import get_user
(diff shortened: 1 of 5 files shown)


leijie-wang:
I took a look at your code and I think everything looks good to me except one small point. The original template use operation names "Add" or "Remove" to represent the action and PolicykitAddUserRole/PolicykitRemoveUserRole as the corresponding action names. Wondering whether we want to keep it consistent

leijie-wang:
As

PR #631, opened 2024-07-04
Add docker

franciclo:
Added docker option for development on local deployment. keeping it very simple.
It introduces two enviroment variables for the database connection and the rabbitmq connection.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.8"
 
 services:
   web:
-    build: .
+    build: ./policykit/
     command: bash -c "python3 manage.py makemigrations &&
       python3 manage.py migrate &&
       python3 manage.py runserver 0.0.0.0:8000"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,11 +3,20 @@ version: "3.8"
 services:
   web:
     build: .
-    command: bash -c "python manage.py migrate && python manage.py runserver 0.0.0.0:8000"
+    command: bash -c "python3 manage.py makemigrations &&
+      python3 manage.py migrate &&
+      python3 manage.py runserver 0.0.0.0:8000"
     volumes:
       - ./policykit:/app
     ports:
       - "8000:8000"
+    environment:
+      - CELERY_BROKER_URL=amqp://guest:guest@rabbit:5672/
+      # - DATABASE_URL=sqlite:///db.sqlite3
+      - DATABASE_URL=postgres://user:password@postgres:5432/policykit
+    depends_on:
+      - rabbit
+      - postgres
 
   celery:
     build: .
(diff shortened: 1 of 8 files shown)


policykit/policykit/settings.py
```diff
@@ -157,11 +157,7 @@
 # https://docs.djangoproject.com/en/3.0/ref/settings/#databases
 
 DATABASES = {
```
saulshanabrook, line 159: It might make sense to add a default for `DATABASE_URL` in the `env` global to preserve the current behavior of reading from the `BASE_DIR`. That way anyone using that key won't be disrupted by this change.

docker-compose.yml
```diff
@@ -0,0 +1,53 @@
+version: "3.8"
+
+services:
+  web:
+    build: .
```
saulshanabrook, line 5:
```suggestion
    build: ./policykit/
```

docker-compose.yml
```diff
@@ -0,0 +1,53 @@
...
+    volumes:
+      - ./policykit:/app
+    ports:
+      - "8000:8000"
+    depends_on:
+      - rabbit
+      - postgres
+
+  celery:
+    build: .
```
saulshanabrook, line 18:
```suggestion
    build: ./policykit/
```

docker-compose.yml
```diff
@@ -0,0 +1,53 @@
...
+    user: django-user
+    command: ["celery", "-A", "policykit", "worker", "--loglevel=info"]
+    volumes:
+      - ./policykit:/app
+    depends_on:
+      - rabbit
+      - postgres
+
+  celerybeat:
+    build: .
```
saulshanabrook, line 28:
```suggestion
    build: ./policykit/
```

docker-compose.yml
```diff
@@ -0,0 +1,53 @@
+version: "3.8"
+
+services:
+  web:
+    build: .
+    command: bash -c "python3 manage.py makemigrations &&
+      python3 manage.py migrate &&
+      python3 manage.py runserver 0.0.0.0:8000"
```
saulshanabrook, line 8: Could move `makemigrations` and `migrate` out of the top level command and instead in the docs to run separately, so that we have more control when they are run and so that the runserver is run as the top level process.

saulshanabrook (commented):
Thanks for putting this together, using Docker Compose should make getting started much easier for new contributors (like myself...)

I was wondering if you would be open to adding instructions to the development docs to describe how to get started with Docker Compose? i.e. just like "Run `docker compose up` and then..." or something.

Also, would it be possible to set it up so that the docker-compose.yml or Dockerfile sets up all the correct env variables properly so that it connects to the database and to rabbit without having to manually edit the `.env` file?

docker-compose.yml
```diff
@@ -0,0 +1,53 @@
+version: "3.8"
+
```
saulshanabrook, line 2:
```suggestion
```

The [version element is now obsolete according to the compose docs](https://docs.docker.com/compose/compose-file/04-version-and-name/#version-top-level-element-obsolete). It will always use the last version now.

policykit/policykit/settings.py
franciclo: Totally, I pushed the changes although I still need to test this since I never got to run the project without docker yet.

docs/source/gettingstarted.rst
```diff
@@ -68,6 +68,29 @@ There is an open issue to support logging in without any third-party platform: `
...
 
 
+Local Development (with docker)
+-----------------
+
+You can also deploy PolicyKit using Docker.
+
+1. Make sure to have Docker and Docker Compose installed.
+
+2. Clone the repo, navigate to the root and create an ``.env`` file as on the previous guide:
```
saulshanabrook, line 78: Would it be possible to have it work so that you didn't have to create a custom `.env` file for Docker but that the variables would automatically be set to pull in the DB and Rabbit connections automatically? I could see this being helpful as we extend the docker compose file with any more service as well, then they should just work without requiring anything on the users part.

Like could be done by adding the env variables directly to the web service in the compose file and [extending it](https://docs.docker.com/compose/multiple-compose-files/extends/#extending-services-within-the-same-file-and-from-another-file) for the celery services, or use a special `.env.docker` file that is pulled into those services?

franciclo: Sorry I missed the last comment on yesterday where you also pointed this.

> then they should just work without requiring anything on the users part.

I didn't solve this in the exact way you suggested, but I think we are getting the same effect anyway. I removed the DB and Rabbit config from `.env.example` since we now have defaults on `settings.py` and added them as environment variables to `docker-compose.yml`. This way we have correct defaults when running both on docker and without it.

saulshanabrook: Thank you!

franciclo:
Right, thats because you should set DEBUG to True as pointed on the docs and in the error. 
Should we hardcode that on `docker-compose.yml`?

saulshanabrook:
Yeah, I feel like it would be nice to be able to start with docker compose without having to edit any of the .env variables manually.

franciclo:
Alright it would be nice to have indeed. You can try it out. 
Did you manage to run the project before? could you confirm this still works without docker?

policykit/policykit/settings.py
```diff
@@ -24,6 +24,8 @@
     ALLOWED_HOSTS=(list, []),
     SERVER_URL=(str, "http://127.0.0.1:8000"),
     LOG_FILE=(str, "debug.log"),
+    DATABASE_URL=(str, "sqlite://" + os.path.join(BASE_DIR, 'db.sqlite3')),
```
saulshanabrook, line 27: I just tried it locally without docker and realized that we need one more `/` to make it an absolute path.

```suggestion
    DATABASE_URL=(str, "sqlite:///" + os.path.join(BASE_DIR, 'db.sqlite3')),
```

saulshanabrook:
Yay, it seems to work now with Docker!

Screenshot 2024-07-06 at 9 28 58 AM


Thank you!

Just left one comment to fix the normal mode as well, then it should be good to go

saulshanabrook (approved):
Looks good to me, I tested it locally and it seems to function with and without docker. 

I wasn't able to sign up on the demo because the slack integration wasn't working, but I think this can be resolved in a follow up PR.

PR #633, opened 2024-08-08
Add support for local slack install with nginx

saulshanabrook:
This PR builds on @franciclo's work in #361 to support login locally using ~~[localtunnel](https://github.com/localtunnel/localtunnel). I choose to use that instead of ngrok because it doesn't require an account.~~

~~However, if we encounter any issues with it, we could switch to ngrok.~~

EDIT: This PR now uses ngrok because I found localtunnel to be unstable. It does require a free ngrok account.

You can manually put in a requested subdomain, and then that will be used to create your public account which you can then create a slack app for.

I tested this and it seems to work. Note that I tried it at first under a slack workspace where I was not an admin and it failed, so make sure to try it with an admin workspace.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,23 +6,27 @@ services:
       - ./policykit:/app
     ports:
       - "8000:8000"
-    env_file: "policykit/policykit/.env"
+    env_file: ".env"
     environment:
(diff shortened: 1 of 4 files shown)


saulshanabrook:
If anyone who has Docker installed could try this out locally to make sure it works for them as well that would be much appreciated!

amyxzhang (approved):

PR #645, opened 2024-09-24
Use different DSN for client

saulshanabrook:
Sentry recommends creating separate projects for client and server. Follow up on #644 

I deployed this already to get error tracking working.


saulshanabrook:
Merging this in ASAP so we can build on it to try to fix the policykit bug

Issue #642, opened 2024-09-11
Error when saving policy in No Code UI


saulshanabrook:
When I try to save a policy in the no code UI I key a server error:

![Screenshot 2024-09-11 at 1 45 10 PM](https://github.com/user-attachments/assets/1aca4c6c-bd9f-47de-9383-04df95cc8004)

```
Traceback (most recent call last):
  File "/environments/policykit/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/environments/policykit/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/environments/policykit/lib/python3.8/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/policykit/policykit/policyengine/views.py", line 986, in create_policy
    new_policy = new_policytemplate.create_policy(user.community.community)
  File "/policykit/policykit/policyengine/models.py", line 1768, in create_policy
    self.create_policy_variables(policy, {})
  File "/policykit/policykit/policyengine/models.py", line 1734, in create_policy_variables
    variable["default_value"] = variable["default"]["value"] if isinstance(variable["default"], dict) else variable["default"]
KeyError: 'default'
```

saulshanabrook:
The line where this error occurred is one where there were some conflicts between what was in the no-code branch and what was added to main. 

The no code branch added this method called `create_policy_variables`:

https://github.com/leijie-wang/policykit/blob/48c5f2992a51957075a9db64abecce3928e83691/policykit/policyengine/models.py#L1691-L1710

Since it branched off, there was a method that was also called `create_policy_variables` that was added:

https://github.com/policykit/policykit/blob/6e99d1545ee4a44434731a49c53647313e9da000/policykit/policyengine/models.py#L1604-L1618


I tried to merge them when I merged master into the branch, but apparently messed up somehow:

https://github.com/policykit/policykit/blob/c7935071d88b41d26a65bf0cf620d89bce05dbf3/policykit/policyengine/models.py#L1722-L1741

saulshanabrook:
Note that currently trying to add a policy in the no code UI like this:

![Screenshot 2024-09-24 at 4 55 59 PM](https://github.com/user-attachments/assets/ae399efc-56e9-40c6-a4b5-d89c2cca9feb)

[...]

PR #673, opened 2024-10-09
Fixed the bug when creating a new policy on the merged interface #642

leijie-wang:
#642 

I did some testing and identified the issue  

the library of policy templates is only reloaded once and for all when we explicitly add a reload parameter to the link "https://policykit.org/no-code/main?reload=true". I have reloaded the templates.
The value of the policy variable has a limit of 100, which is too short for some very long messages. I have increased its value.
I will later upload the new branch to this repo.


amyxzhang:
Approving and merging this but can you deploy and carry out the migration in the prod db @leijie-wang?

amyxzhang (approved):

PR #675, opened 2024-10-09
Fix sphinx immaterial version conflicts

saulshanabrook:
This is a follow up on #581 which fixes some version conflicts caused by installing sphinx immaterial. It removes the sphinx rtd theme, since it is no longer used, and also upgrades the sphinx version so that it is compatible with sphinx immaterial.

Closes https://github.com/policykit/policykit/issues/674

### DIFF
--- a/policykit/requirements.txt
+++ b/policykit/requirements.txt
@@ -1,4 +1,4 @@
-sphinx-immaterial==0.11.2
+alabaster==0.7.13
 amqp==2.5.2
 appdirs==1.4.4
 appnope==0.1.0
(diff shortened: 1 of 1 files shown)


policykit/requirements.txt
```diff
@@ -96,6 +105,3 @@ wrapt==1.12.1
 zipp==2.0.1
 zope.event==4.5.0
 zope.interface==5.1.2
-
-django-pattern-library
-djangorestframework
```
saulshanabrook, line 101: I ran pip freeze so these got moved to explicit pins

saulshanabrook:
I opened https://github.com/policykit/policykit/issues/676 to track more generally upgrading all of our packages to avoid this sort of thing in the future.

amyxzhang (approved):

PR #678, opened 2024-10-12
Fixed the bug when creating a new policy on the merged interface #642

leijie-wang:
While the previous fix works for most policies, I forgot to increase the limit for the default value as well.


saulshanabrook (approved):
Thanks!

PR #679, opened 2024-10-22
Turn on atomic requests

saulshanabrook:
This turns on atomic requests, so that if some part of a response fails, all changes to the DB will be reverted.

This should help https://github.com/policykit/policykit/issues/642 kind of bug from corrupting the database.


saulshanabrook:
I also added a missing migration from #678

amyxzhang (approved):

PR #686, opened 2024-11-06
Allow users to edit a policy created from nocode interface later. 

leijie-wang:


saulshanabrook (commented):
Hey, thanks for working on this! From a quick glance, one thing I notice is I am unsure about changing/deleting old migrations... I worry that will be hard to, well, migrate too!

policykit/templates/no-code/main.html
```diff
@@ -237,6 +237,10 @@ <h5 class="modal-title prompt small">Discard the Policy</h5>
 {% block step_scripts %}
 
 <script>
+
+    const policytemplate = JSON.parse('{{policytemplate | safe}}');
+    console.log('policytemplate', policytemplate)
```
amyxzhang, line 242: do you want to keep this log statement in?

policykit/templates/no-code/main.html
```diff
@@ -607,55 +633,64 @@ <h5 class="modal-title prompt small">Discard the Policy</h5>
...
+            if(set_variable) variable.value = set_variable.value;
             addVariableInputBox(variable=variable, id=variable.name, parentDiv=procedure_div, options=entities, datalist=global_datalist, type="procedure")
         })
 
         
     }
         
-    const showTransformerVariables = (transformer_value, app) => {
+    const showTransformerVariables = (transformer_value, app, applied_transformer={}) => {
+        console.log('show transformer variables', applied_transformer)
```
amyxzhang, line 657: same question

policykit/templates/no-code/main.html
```diff
@@ -607,55 +633,64 @@ <h5 class="modal-title prompt small">Discard the Policy</h5>
...
 
-        let selected_transformer = {...transformers[app].find(item => item.value == transformer_value)};
+        let selected_transformer = deepcopy(transformers[app].find(item => item.value == transformer_value));
         selected_transformers.push(selected_transformer);
         
     
         let transformer_div = add_section_header("transformer", selected_transformer);
         $("#design_transformer-formview > :last-child").before(transformer_div);
         selected_transformer["variables"].forEach((variable, index) => {
+            console.log('variable', variable)
```
amyxzhang, line 669: same question

policykit/integrations/slack/models.py
```diff
@@ -228,7 +228,7 @@ def join_conversation(self, channel, users):
 class SlackPostMessage(GovernableAction):
     ACTION = "chat.postMessage"
     AUTH = "admin_bot"
-    EXECUTE_PARAMETERS = ["text", "channel", "thread"]
+    EXECUTE_PARAMETERS = ["text", "channel"]
```
amyxzhang, line 231: why did we remove this?

policykit/policyengine/models.py
```diff
@@ -983,6 +983,14 @@ class Policy(models.Model):
     modified_at = models.DateTimeField(auto_now=True)
     """Datetime object representing the last time the policy was modified."""
 
+    policy_template = models.OneToOneField(
```
amyxzhang, line 986: does this require a migration file?

policykit/templates/no-code/main.html
leijie-wang: removed such logs

policykit/integrations/slack/models.py
leijie-wang: I recalled that I made this change because of a bug I found in the deployment (though forgot exactly what the bug was). For reference, this variable is only used in old policykit logic when the user wants to propose a new action directly on the policykit website. However, the SlackPostMessage does not have a thread field. In the no-code interface, users could still have the thread field because another variable 'EXECUTE_VARIABLES' right after this line determines this.

policykit/templates/no-code/main.html
amyxzhang: sorry did you add a commit that removed them? I still see them all in the PR.

policykit/policyengine/models.py
leijie-wang: Yes; I did try to migrate it but could not do it on my instance because the newly added code requires python3.10 which is not that compatible with my ubuntu instance 18.04 --- a huge rabbit hole. wondering whether we could migrate on the production instance after merging the code

leijie-wang: created  the new migration file

policykit/templates/no-code/main.html
leijie-wang: yeah, I kind of want to because it represents the middle product and can be very useful for debugging

amyxzhang (approved):

policykit/policyengine/views.py
```diff
@@ -298,42 +298,51 @@ def editor(request):
...
-
-    data = {
-        'user': get_user(request),
-        'type': kind.capitalize(),
-        'operation': operation,
-        'actions': actions.items(),
-        'autocompletes': json.dumps(autocompletes),
-        'base': base
-    }
+    if not policy.policy_template:
```
saulshanabrook, line 301: I am trying this locally with a URL like `https://loyal-guinea-instantly.ngrok-free.app/no-/editor/?policy=1` and see that it doesn't render the new UI because there is not template for this policy.

Doesn't this mean that any policies not created in the new UI wont have a policy template and so wont be editable with the new UI?

saulshanabrook (commented):
I can't seem to get this working locally...

Is there a way to load existing policies with the new no-code editor?

If not, to test it, you have to make a new policy and then editing it right? I was having a hard time even making a new policy. I was getting an error `PolicyTemplate matching query does not exist.` when trying to save a policy with this PR.

policykit/policyengine/views.py
leijie-wang: Yes, otherwise I do not know which component to render if it is not created with the no-code UI.

leijie-wang:
> I can't seem to get this working locally...
> 
> Is there a way to load existing policies with the new no-code editor?
> 
> If not, to test it, you have to make a new policy and then editing it right? I was having a hard time even making a new policy. I was getting an error `PolicyTemplate matching query does not exist.` when trying to save a policy with this PR.

leijie-wang:
> I can't seem to get this working locally...
> 
> Is there a way to load existing policies with the new no-code editor?
> 
> If not, to test it, you have to make a new policy and then editing it right? I was having a hard time even making a new policy. I was getting an error `PolicyTemplate matching query does not exist.` when trying to save a policy with this PR.

Yes, you have to first load existing component templates with reload=true parameter when entering the no-code interface, for instance, 'http://prototype.policykit.org/no-code/main?reload=true'

policykit/policyengine/views.py
amyxzhang: Hm, so what happens right now with old policies? does the page just break? Or does it route to the old UI? 

ideally we would like to be able to view all policies with the same UI. Can you start a new PR for showing the code instead of the form UI? then we can address this in an upcoming PR instead of here.

leijie-wang: right now it routes to the old UI; okay, I will create a new PR and try to implement this feature this month
.

PR #690, opened 2024-12-27
Fix slow loading on homepage

saulshanabrook:
This PR fixes the slow loading on the homepage. It now takes only a few seconds to load the API locally, instead of previously it would take minutes, if it finished at all.

The main fix was using `selected_related` to prefetch some proposal data, otherwise an exponential number of queries would be run.

To help debug this issue, I created a custom management command that runs the API response to produce the dashboard and logs how many queries were executed.


amyxzhang:
Hurrah!!

amyxzhang (approved):

PR #693, opened 2025-01-13
Go through all pending proposals even if one fails

saulshanabrook:
This PR fixes a bug where if one proposal would fail to get a community name then then loop would stop running. Instead, we can just log the error and keep going.

I am not sure why this community name fails, but I noticed it was happening frequently in production.

### DIFF
--- a/policykit/policyengine/tasks.py
+++ b/policykit/policyengine/tasks.py
@@ -21,7 +21,11 @@ def evaluate_pending_proposals():
     #logger.debug("Running evaluate_pending_proposals:" + str(len(pending_proposals)))
 
     for proposal in pending_proposals:
-        community_name = proposal.action.community.community_name
+        try:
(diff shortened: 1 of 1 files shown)


saulshanabrook:
Wow, thanks Sentry! Yeah that first one is the error that this fixes: https://policykit.sentry.io/issues/6012782096/?project=4507970394980352&referrer=github-open-pr-bot

The second was a temporary error I introduced in prod when trying out a mistaken version of this fix.

amyxzhang:
wow very interesting bot feature! LGTM

amyxzhang (approved):
I'm not sure why there was an action where we don't have an associated community name, at some point it would be good to look into it.

saulshanabrook:
I opened https://github.com/policykit/policykit/issues/695 to track the bug

PR #696, opened 2025-02-18
Adds Proposals to the main dashboard

saulshanabrook:
This PR adds a right column to the dashboard with "proposals" that have had actions taken on them:

Screenshot 2025-02-18 at 11 56 28 AM

It is based roughly on the "Governance History" section on the mockups:

Screenshot 2025-02-18 at 12 06 23 PM

I displayed "proposals" instead of "governable actions", because they were already serialized (so thinking that whoever worked on it last intended them to be used) and because "governable actions" dont seem to have a date/time associated with them AFAIK. So I wasn't sure how to display them.

I did set up serialization for them and could add them if we wish! That would I believe get us closer to the semantics of the figma.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ postgres_data
 celerybeat.pid
 celerybeat-schedule
 log
+models.svg
+models.dot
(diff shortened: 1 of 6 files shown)


saulshanabrook:
Talked about in meeting - Wanted to separate pending proposal above, below is completed proposals, so its easier to see what is still in progress.

saulshanabrook:
Try to either align action name or string to get something better for describing what happened.

amyxzhang (approved):

PR #698, opened 2025-03-18
Change proposals list to show action type

saulshanabrook:
Follow up from #696 to show action type along with each proposal

Screenshot 2025-03-18 at 3 03 52 PM

Screenshot 2025-03-18 at 3 03 59 PM

Uses `ACTION_NAME` and falls back on `str` if that is not defined.

### DIFF
--- a/frontend/src/dashboard.tsx
+++ b/frontend/src/dashboard.tsx
@@ -19,6 +19,7 @@ type PolicySummary = {
 type ActionSummary = {
   id: number;
   action_type: string;
+  description: string;
 };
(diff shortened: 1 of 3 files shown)


saulshanabrook:
This also now separates the completed from still pending proposals:

Screenshot 2025-03-18 at 3 37 01 PM

saulshanabrook:
Switch logic to use is vote closed - nothing currently open. and update post message str to remove id

amyxzhang (approved):

